PR #2, opened 2021-11-19
Documentation outline

gelzinyte:
Documentation moved from the old Wiki. Build with `make html` (etc) from `workflow/docs` & `open build/html/index.html`.


bernstei:
If I can confirm that it works, any reason not to merge immediately?

bernstei:
I just did a fresh "pip install Sphinx", and I'm getting
```
tin 1023 : make html
Running Sphinx v4.3.0
making output directory... done

Extension error:
Could not import extension sphinx_book_theme (exception: No module named 'bs4')
make: *** [Makefile:20: html] Error 2
```

gelzinyte:
Auto-documentation is still missing, for which issue #3 should get resolved. In the same go, I think it'd be neat to set up github workflow to automatically publish to  https://libatoms.github.io/workflow. 

Yes - `pip install sphinx-book-theme`, it's not one of the defaults (and easy to change)

bernstei:
OK, `sphinx-book-therem` works to get started, but we should beware that it seems to require sphinx < 4, so it downgraded my other sphinx, and that's not a  problem in and of itself, but it may mean that support for `sphinx-book-theme` might go away at some point.

bernstei:
I'm happy to merge.  No reason to keep the branches separate longer than we need to.  @gelzinyte Do you want to confirm that what I did didn't actually break it for you first?

PR #8, opened 2021-12-08
clean up iterable_loop, including decorator and outdated wfl.parallelization

### LINKED ISSUE #7: get rid of wfl.parallelization
`wfl.parallelization` has been entirely (I think) supplanted by `wfl.pipeline`.  A few things still use it, and those should be redone.

bernstei:
Make a decorator form of iterable_loop() called iloop(), applied using functools.partial (see wfl/pipeline/base.py)

Get rid of wfl/parallelization.py, only used by wfl.generate_configs.collision.
Try to make collision use new iloop(), but untested so far.

closes #7


wfl/pipeline/__init__.py
```diff
@@ -1,2 +1,5 @@
-from .base import iterable_loop
+from .base import iterable_loop, iloop, iloop_docstring_pre, iloop_docstring_post
 assert iterable_loop
+assert iloop
+assert iloop_docstring_pre
```
gelzinyte, line 4: Why the asserts?

bernstei: I don't know.  They were there for `iterable_loop`, so I assumed someone added them for a reason.  Let me see if I can figure out why.

bernstei: It was in a commit labeled "Lots of flake8 corrections in wfl/qurefun and wfl/pipeline", so I suspect it gave a warning about an unused import or something like that.

PR #13, opened 2022-01-03
option to only return only converged configs

gelzinyte:


wfl/generate_configs/minim.py
```diff
@@ -201,16 +189,41 @@ def process_step():
...
-        # Perhaps status should be set after resampling?
-        traj = _resample_traj(traj, traj_equispaced_n)
 
         if update_config_type:
             # save config_type
             for at0 in traj:
                 config_type_append(at0, at0.info['minim_config_type'])
 
-        all_trajs.append(traj)
+        if traj_subselect is not None:
```
bernstei, line 198: Probably simpler to have `_resample_traj` deal with `None`, given all the other `if elseif` we'll need for various resampling criteria.

wfl/generate_configs/minim.py
```diff
@@ -201,16 +189,41 @@ def process_step():
...
+            # Note that if resampling doesn't include original last config, later
+            # steps won't be able to identify those configs as the (perhaps unconverged) minima.
+            # Perhaps status should be set after resampling?
+            traj = _resample_traj(traj, subselect=traj_subselect)
+
+        if len(traj) > 0:
```
bernstei, line 204: Have you thought about what'll happen when these fail?  Under at least some circumstances the `ConfigSet_out` cleverness tries to match inputs to outputs, and I think it'll behave badly where there's no corresponding output.  Maybe that's OK, and you just can't do the outputs matching inputs in that situation, but it's worth checking. Maybe we should append an empty list, rather than appending nothing?

Also, what happens if _every_ minim fails, so there's no output at all.

gelzinyte: Good points! `pool.do_in_pool()` actually handles `None`s ([L51](https://github.com/libAtoms/workflow/blob/e654f6fe4a7611a655730877b76114443ba8b0e8/wfl/pipeline/pool.py#L51) & [L155](https://github.com/libAtoms/workflow/blob/e654f6fe4a7611a655730877b76114443ba8b0e8/wfl/pipeline/pool.py#L155)), so let me fix `minim.run_op()` to return `None` for each failed minimisation.

wfl/generate_configs/minim.py
gelzinyte: changed in the last commit

wfl/generate_configs/minim.py
gelzinyte: Fixed in the last commit.

tests/test_minim.py
```diff
@@ -152,3 +151,42 @@ def test_relax_fixed_vol(cu_slab):
...
+
+    cu_slab_optimised = cu_slab.copy()
+    cu_slab_optimised.set_positions(expected_relaxed_positions_constant_pressure)
+
+    inputs = ConfigSet_in(input_configs = [cu_slab, cu_slab_optimised] )
+
+    # returns full optimisation trajectories
+    atoms_opt = minim.run_op(inputs, calc, fmax=1e-2, precon=None,
+                          logfile='-', verbose=True, pressure=-1.1215, 
+                          steps=2, traj_subselect=None)
```
bernstei, line 168: What do you think about changing the name of this variable OR the function that's called, so they either both use "subselect" or "resample"?

bernstei (approved):
I'm happy with it as is, with just one optional additional change (see comment).

tests/test_minim.py
gelzinyte: I like consistency, thanks for pointing this out. I went with "resample", I think that is _slightly_ more intuitive.

gelzinyte:
Thanks! Is it okay to merge or should I wait for the ace patch?

bernstei:
> Thanks! Is it okay to merge or should I wait for the ace patch?

I'd prefer to wait for the patch.  Should only be a few days, and I'll fix any conflicts that arise.

PR #14, opened 2022-01-03
Path support for ConfigSet_in & ConfigSet_out

gelzinyte:


wfl/configset.py
```diff
@@ -397,18 +397,24 @@ def __init__(self, abcd_conn=None, set_tags=None,
...
-                            raise ValueError(f'Got type {type(output_files)} of length {len(output_files)} other '
-                                             f'than 1 or content type other than str, cannot map to it')
+                            raise ValueError(f'Got `output_files` of type {type(output_files)} and length {len(output_files)}, '
+                                             f'which either has more than one entry or its content is not one of: '
+                                             f'str, dict, pathlib.Path')
                     except Exception as e:
                         traceback.print_exc()
-                        raise RuntimeError(f'Got output_files type {type(output_files)} other than iterable with '
-                                           f'1 str item, str, or dict')
+                        raise RuntimeError(f'Got output_files of type {type(output_files)} wich is not one of: '
```
bernstei, line 416: typo "wich"

bernstei (commented):
Basically OK, but I don't want to merge yet, until I do my ACE fit patch, because it's going to be slightly difficult to apply if other changes happen before I do.  Should only be another few days.

wfl/configset.py
gelzinyte: thanks!

gelzinyte:
> Basically OK, but I don't want to merge yet, until I do my ACE fit patch, because it's going to be slightly difficult to apply if other changes happen before I do. Should only be another few days.

Good with me! I'll also add `Path` support to `ConfigSet_in`

bernstei:
It might also be nice, while we're at it, to get rid of all the "os.path" calls and replace them with corresponding Pathlib syntax, probably mostly things like "os.path.join(a, b)" -> "Path(a) / b".  Or maybe even storing file paths as `Path`s instead of strings internally.

gelzinyte:
changing to `Path`s sounds good and shouldn't be tricky to change, let me. 

I'm looking at `ConfigSet_in` and `in_memory()` (L252). Wouldn't `input_configs=list(self)` also work for when `self.input_configs is not None`? 

```
    def in_memory(self):
        """Create a ConfigSet_in containing the same configs, but in memory (i.e. as ``self.input_configs``)

        Returns
        -------
        ConfigSet_in
        """

        if self.input_configs is not None:
            return ConfigSet_in(input_configs=self.input_configs)
        else:
            return ConfigSet_in(input_configs=list(self))
```

wfl/configset.py
```diff
@@ -394,22 +403,34 @@ def __init__(self, abcd_conn=None, set_tags=None,
                 print('writing to ABCD')
         elif output_files is not None:
             if isinstance(output_files, dict):
-                self.output_files = [os.path.join(file_root, fout) for fout in output_files.values()]
-                self.output_files_map = lambda fin: os.path.join(file_root, output_files.get(fin))
+                for key, val in output_files.items():
+                    if isinstance(val, str):
+                        output_files[key] = Path(val)
```
gelzinyte, line 408: Any better way to convert `output_files` entries to Paths?

wfl/configset.py
```diff
@@ -585,3 +612,20 @@ def __str__(self):
         if self.abcd is not None:
             s += _fmt('ABCD connection', str(self.abcd))
         return s
+
+def _from_dict_str_to_dict_Path(files_dict):
+    """ converts all strings in the dictionary  to Paths"""
+
```
gelzinyte, line 618: Not sure if this is the best way?

gelzinyte:
@bernstei what do you think?

wfl/configset.py
```diff
@@ -50,8 +50,8 @@ class ConfigSet_in:
         path to prepend to every file name
     input_files: str / iterable(str) / 2-tuple(str) / iterable(2-tuple(str))
         Input file name globs, optionally as 2-tuples of (filename_glob, index).
-        The latter only works with more than one tuple
-        index here overrides separate default_index arg
+        The latter only works with more than one tuple index here overrides
```
bernstei, line 53: line breaks were meaningful before.  Need "." after "one tuple".

wfl/configset.py
```diff
@@ -104,6 +104,9 @@ def __init__(self, abcd_conn=None,
         elif input_files is not None:
             # fix up files
             self.input_files = []
+            if isinstance(input_files, Path):
+                # Path.glob() below takes string as argument
+                input_files = str(input_files)
```
bernstei, line 109: If we convert `file_root` from str **to** Path before storing it, why are we doing the opposite here?  Is there a reason not to store all files/paths as Path objects, and converting to str when necessary?  We will need to do it when passing to `glob.glob()`, but `ase.io.read` can already accept Paths.

wfl/configset.py
bernstei: I'm not sure if it's exactly "better", but it's certainly more compact to do
```
    output_files = {Path(k): Path(v) for k, v in output_files.items() }
```
since `Path(Path)` and `Path(str)` both just result in a `Path`.

wfl/fit/gap_simple.py
```diff
@@ -112,9 +112,9 @@ def run_gap_fit(fitting_configs, fitting_dict, stdout_file, gap_fit_exec="gap_fi
     if not fitting_configs_filename:
         fitting_configs_scratch_filename = fitting_configs.to_scratch_file('_GAP_fitting_configs.xyz')
         fitting_configs_filename = fitting_configs_scratch_filename
-
+    
     # kwargs overwrite the fitting_dict given
-    use_fitting_dict = dict(fitting_dict, atoms_filename=fitting_configs_filename, **kwargs)
+    use_fitting_dict = dict(fitting_dict, atoms_filename=str(fitting_configs_filename), **kwargs)
```
bernstei, line 117: If we do decide to go with always storing `Path` internally, maybe `dict_to_gap_fit_string()` is the place where it should convert from Path to str.

wfl/configset.py
gelzinyte: of course; reverted back.

wfl/configset.py
gelzinyte: Yes, this is redundant. Forgot to remove it after changing to `glob.glob()` below.

bernstei:
@gelzinyte I'm confused - I thought I got an email with a comment from you mentioning `glob.glob`, but I don't see it here.  Am I missing something?

wfl/fit/gap_simple.py
gelzinyte: Everything gets converted to string by `ase.io.extxyz.key_val_dict_to_str`, which doesn't recognise `Path` and omits the entry altogether. I'll see about suggesting a fix in ase.

gelzinyte:
That's probably in the second resolved conversation (near L109 in configset.py) from your latest review?

wfl/fit/gap_simple.py
bernstei: Oh, yes.  Probably not worth trying to get a change into ase.io, since we're planning to switch to an entirely different extxyz reader soon anyway.

bernstei:
I guess it's OK now, except I also want this to wait until after the big ACE patch.

wfl/fit/gap_simple.py
gelzinyte: Ah `pathlib` is not a standard module, so adding a two-line check to `extxyz.py` would need to introduce a new dependency to ase - I don't think that's an option (?).  I'll add a function to `dict_to_gap_fit_string()`.

bernstei:
> On Jan 5, 2022, at 1:12 PM, gelzinyte ***@***.***> wrote:
> 
> 
> @gelzinyte commented on this pull request.
> 
> In wfl/fit/gap_simple.py <https://github.com/libAtoms/workflow/pull/14#discussion_r779034978>:
> 
> >      # kwargs overwrite the fitting_dict given
> -    use_fitting_dict = dict(fitting_dict, atoms_filename=fitting_configs_filename, **kwargs)
> +    use_fitting_dict = dict(fitting_dict, atoms_filename=str(fitting_configs_filename), **kwargs)
> Ah pathlib is not a standard module, so adding a two-line check to extxyz.py would need to introduce a new dependency to ase - I don't think that's an option (?). I'll add a function to dict_to_gap_fit_string().
> 
I thought that ase.io <http://ase.io/> already supports pathlib.  Maybe they just convert to string, and that just works automatically if they receive a Path.

wfl/fit/gap_simple.py
gelzinyte: didn't see your message above - ok!

gelzinyte:
> I thought that ase.io <http://ase.io/> already supports pathlib.  Maybe they just convert to string, and that just works automatically if they receive a Path.

I was judging by `setup.py` contents, but installing ase in fresh environment installs pathlib too after all. In any case, I'll leave it as is. 

> I guess it's OK now, except I also want this to wait until after the big ACE patch.

Ok with me, thanks for the feedback!

PR #15, opened 2022-01-08
Rollup of various changes related to new ACE fitting functionality

bernstei:
Actual fitting is done using a julia script in wfl/scripts.  In addition to new fitting function in wfl/fit/ace.py, also moved database modification (pre fit) and fitting/testing error (post fit) out of fit/gap_multistage and into `gap_rss_iter_fit.py`.

New function to generate phonon and 3rd order force constant displacements with phonopy/phono3py.

Replace single_atom with isolated_atom for special config type.

A few other little changes here and there.

@gelzinyte do you want to look it over before we merge, assuming the tests all pass?


wfl/calculators/generic.py
```diff
@@ -52,7 +53,9 @@ def run_op(atoms, calculator, properties=None, output_prefix='_auto_', verbose=F
         at.calc = calculator
         calculation_succeeded = False
         try:
-            at.calc.calculate(at, properties=properties)
+            # explicitly pass system_changes=None because some calculators, e.g. ace.ACECalculator,
```
gelzinyte, line 56: typo

wfl/fit/ace.py
```diff
@@ -0,0 +1,299 @@
...
+
+    ace_file_base = str(run_dir / ACE_name)
+
+    def _read_size():
+        with open(ace_file_base + '.size') as fin:
+            fields = fin.readline().strip().split()
+            return (int(fields[3]), int(fields[5]))
+
+    # figure out whether fit calculations are needed
+    need_to_fit = True
```
gelzinyte, line 85: I don't like having the `need_to_fit` flag, could it do without? 

For example

```
    # figure out whether fit calculations are needed
    if skip_if_present:
        # see if we can open and parse each output file
        try:
            _... not showing some code ..._ 
                        warnings.warn('Cannot parse ACE file with suffix '+suffix)
            if dry_run:     # previous "dry_run and not need_to_fit" would have only been True if an exception hasn't been raised, so can do here
                return size
            else:
                return ace_file_base  # "not need to fit" would have only been True if exception hasn't been raised, do here
        except:
                pass   # couldn't read file, need to fit.    

    if remote_info is not None and remote_info != '_IGNORE': 
        ...
```

wfl/fit/ace.py
```diff
@@ -0,0 +1,299 @@
...
+                        _ = json.load(fin)
+                    # yace files are apparently not parseable by python yaml module
+                    # elif suffix == '.yace' or suffix == '.yaml':
+                        # _ = yaml.safe_load(fin)
+                    elif suffix == '.size':
+                        # dry run results
+                        size = _read_size()
+                    else:
+                        warnings.warn('Cannot parse ACE file with suffix '+suffix)
+        except:
```
gelzinyte, line 108: what type of exception is expected?

wfl/fit/ace.py
```diff
@@ -0,0 +1,299 @@
...
+        # the code below needs to know an unfortunate amount about the inner workings of ace_fit.jl
+        # here we rely on knowledge of the --outfile_base
+        use_params['outfile_base'] = ace_file_base
+        use_params['outfile_format'] = formats
+
+        use_params['key'] = [ [k, ref_property_prefix + v] for k,v in [('E', 'energy'), ('F', 'forces'), ('V', 'virial')] ]
+        use_params['_repeat_key'] = True
+
+        # configs need to be in memory so they can be modified with stress -> virial
+        fitting_configs = list(fitting_configs)
```
gelzinyte, line 168: why not `ConfigSet_in.in_memory()`?

gelzinyte: Maybe not much use after all - I initially thought having a configsetin would slightly simplify reading/writing ~10 lines below

wfl/fit/gap_multistage.py
```diff
@@ -175,23 +156,21 @@ def fit(fitting_configs, GAP_name, params, ref_property_prefix='REF_',
 
     if remote_info is not None and remote_info != '_IGNORE':
         # check if we can skip
-        need_to_recalc = True
+        need_to_fit = True
```
gelzinyte, line 159: Similarly to `ace.py`, could move the for loop into try-except, move returns from lines ~200 into the `try` part, so that exception only happens if the fit is needed. I think that will make the code clearer.

wfl/fit/gap_multistage.py
```diff
@@ -175,23 +156,21 @@ def fit(fitting_configs, GAP_name, params, ref_property_prefix='REF_',
...
         if skip_if_present:
-            need_to_recalc = False
+            need_to_fit = False
             for final_GAPfile in final_GAPfiles:
                 try:
                     cElementTree.parse(final_GAPfile)
                 except:
```
gelzinyte, line 165: What exception is expected?

gelzinyte:
The tests aren't passing - looks like phonopy is missing from `.github/workflows/python-package-conda.yml`?

I've left some small comments, but otherwise looks good. I'll extend `ace_fit.jl` with what I'll use in another pull request.

What do you think about making a proper (small) julia package instead of ace_fit.jl? It'd make it more modular, testable and easier to develop.

wfl/fit/ace.py
bernstei: It was originally needed when the code also did the fitting/testing error calculation, and there were two flags, one for the actual fit, one for error calculation.  Let me think about whether it can be removed now.

wfl/fit/ace.py
bernstei: I think it's just copied from code that predates that function.  Let me think if there's any reason to prefer one or the other syntax.

wfl/calculators/generic.py
bernstei: fixed in 

> The tests aren't passing - looks like phonopy is missing from `.github/workflows/python-package-conda.yml`?

Yes, I'll fix that.
 
> I've left some small comments, but otherwise looks good. I'll extend `ace_fit.jl` with what I'll use in another pull request.
> 
> What do you think about making a proper (small) julia package instead of ace_fit.jl? It'd make it more modular, testable and easier to develop.

I'm happy either way.  My initial thought was just the simplest thing, analogous to `gap_fit` (although not using the same legacy command line syntax), but if you or someone else likes another approach, that's find by me.

bernstei:
> The tests aren't passing - looks like phonopy is missing from `.github/workflows/python-package-conda.yml`?

Yes, I'll fix that.
 
> I've left some small comments, but otherwise looks good. I'll extend `ace_fit.jl` with what I'll use in another pull request.
> 
> What do you think about making a proper (small) julia package instead of ace_fit.jl? It'd make it more modular, testable and easier to develop.

I'm happy either way.  My initial thought was just the simplest thing, analogous to `gap_fit` (although not using the same legacy command line syntax), but if you or someone else likes another approach, that's find by me.

wfl/fit/gap_multistage.py
bernstei: I agree.  Refactored fairly substantially in 16cc346.

wfl/fit/ace.py
bernstei: fixed in 16cc346

wfl/fit/ace.py
bernstei: refactored to clean up in 16cc346

wfl/fit/ace.py
bernstei: Below we pass that variable to `ase.io.write()`, and I thought it's safest to do it as a list. Comment explains this in 683738a.

I guess the real question is whether `to_memory()` is every worth it, but that's a matter for a different PR.

wfl/fit/gap_multistage.py
bernstei: Should be fixed in b5d31ff.  Can you take a look?

gelzinyte: Looks good to me!

PR #18, opened 2022-01-11
set timeout for remote jobs, and fix phonopy displacement generation

bernstei:
Add support for passing timeout to `ExPyRe.get_results()`, rather than the hardwired default of 3600 s.  Also work around phono3py bug (perhaps now fixed in https://github.com/phonopy/phono3py/pull/65) by always using phonopy (rather than phono3py) for harmonic force constant (phonon) displacements.

### DIFF
--- a/tests/test_phonopy.py
+++ b/tests/test_phonopy.py
@@ -21,10 +21,14 @@ def test_phonopy(tmp_path):
 
     pert = list(pert)
 
-    assert len(pert) == 1*2 + 6*2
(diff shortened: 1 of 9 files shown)


gelzinyte (approved):
Looks good as far as I can tell - I haven't used phonopy & related, but the code looks good to me.

bernstei:
We'll see if the phonopy stuff turns out useful in the long term.  Right now I'm trying it for crystals with symmetry breaking structure phase transitions, which are conventionally described by anharmonic phonon models.

PR #20, opened 2022-01-12
Fix bugs in handling of remote_info for local and remote jobs that do fitting

bernstei:
Fix and standardize handling of remote_info in ace, gap_simple, and gap_multistage fitting wrappers

[ADDED LATER]
Fix some failing tests (not related to initial `remote_info` focus)
Fix some deprecated syntax, mainly spglib.

### DIFF
--- a/tests/test_configset_in.py
+++ b/tests/test_configset_in.py
@@ -33,12 +33,12 @@ def test_configset_from_ConfigSets(tmp_path):
     assert len([at for at in c_ats]) == 4
 
 
-@pytest.mark.xfail(reason="should fail creating ConfigSet_in from mix of ConfigSet_ins using input files and input configs")
 def test_configset_from_ConfigSets_mismatched(tmp_path):
     ats = [Atoms('H'), Atoms('C')]
     ase.io.write(tmp_path / 'at1.xyz', ats)
 
-    c_ats = ConfigSet_in(input_configsets=[ConfigSet_in(input_files=str(tmp_path / 'at1.xyz')), ConfigSet_in(input_configs=ats)])
+    with pytest.raises(RuntimeError):
+        c_ats = ConfigSet_in(input_configsets=[ConfigSet_in(input_files=str(tmp_path / 'at1.xyz')), ConfigSet_in(input_configs=ats)])
 
 
 def test_merge():
--- a/tests/test_configset_in.py
+++ b/tests/test_configset_in.py
@@ -33,7 +33,7 @@ def test_configset_from_ConfigSets(tmp_path):
     assert len([at for at in c_ats]) == 4
 
 
-@pytest.mark.xfail()
+@pytest.mark.xfail(reason="should fail creating ConfigSet_in from mix of ConfigSet_ins using input files and input configs")
 def test_configset_from_ConfigSets_mismatched(tmp_path):
     ats = [Atoms('H'), Atoms('C')]
     ase.io.write(tmp_path / 'at1.xyz', ats)
--- a/README.md
+++ b/README.md
@@ -53,18 +53,22 @@ abcd.get_atoms(output_tags)
 
 Operations that have been wrapped in `iterable_loop` can be split into independent jobs with minimal
 coding.  A `config.json` file must describe the available queuing sytem resources (see
-[expyre README](wfl/expyre/README.md)), and a JSON file (content or path in
+[expyre README](https://github.com/libAtoms/ExPyRe#readme)), and a JSON file (content or path in
 env var `WFL_AUTOPARA_REMOTEINFO`) describes the resources needed by any `iterable_loop` call that
 should be executed this way.  Any remote machine to be used requires that the `wfl` python
 module be installed.  If needed, commands needed to make this module available (e.g. setting `PYTHONPATH`)
 can be set on a per-machine basis in the `config.json` file mentioned below.
 
-In addition, `wfl.fit.gap_simple` and `wfl.fit.gap_multistage` have been wrapped, as a single
-job each.  The simple fit is controlled by the `WFL_GAP_SIMPLE_FIT_REMOTEINFO` env var.  Setting
+In addition, `wfl.fit.gap_simple`, `wfl.fit.gap_multistage`, and `wfl.fit.ace` have been wrapped, as a single
+job each.  The GAP simple fit is controlled by the `WFL_GAP_SIMPLE_FIT_REMOTEINFO` env var.  Setting
 this variable will also lead to the multistage fit submitting each simple fit as its own job.
 In addition, the multistage fit can be turned into a single job containing all the stages
 with the `WFL_GAP_MULTISTAGE_FIT_REMOTEINFO` env var.  In principle, doing each simple fit
 as its own job could enable running committee fits in parallel, but that is not supported right now.
+The env var `WFL_ACE_FIT_REMOTEINFO` is used for ACE fits.
+
+[NOTE: now that the multistage fit does very little other than the repeated simple fitting, does
+it need its own level of remote job execution]
 
 ### Example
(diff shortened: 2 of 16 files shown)


bernstei:
I think the need for this PR points out a need for better testing, which I'll try to do, but it won't be automated, because it requires running a remote job which can't be done in CI.

bernstei:
One thing we _could_ do is require that there be a checked in file with the results of a complete test, and that there are no unexpected failures. It would mean that for any PR to be merged _someone_ who can do such as test will have to grab it, run the test, and push an update version of that file. Is that only me, or is anyone else available?  Ideally this would be a singly machine where you could run the tasks locally as well as submitted to a queuing system (even better would be a queuing system on a _remote_ machine, which needs to be accessed with passwordless ssh).

bernstei:
@gelzinyte I'm trying to make sure that tests pass.  I'm getting a failure having to do with your new subselect last_converged option to minim.  Can you explain why the very last test in `test_minim.py` is supposed to work?

bernstei:
I guess I understand - the iterator is just supposed to skip groups that contain None. The real problem is that the earlier minims modify the Atoms structures in memory, so they _both_ converge even in the 3rd call.  Fixed now.  Take a look at ny next push in this PR.

bernstei:
I think everything is as fixed as it's going to get.  The QE test has a hard-coded calculation which does not converge using the default SCF threshold, and when I loosen it the values it returns are not equal to the hard-coded expected ones.  It's marked XFAIL for now, but we should dig into the problem.  Perhaps before it passed because Espresso did not raise an exception even if SCF convergence ran out of iterations?

bernstei:
@gelzinyte  Do you want to go over this one before I merge?

gelzinyte:
Sorry didn't get to this either - let me take a look before you do though

tests/test_configset_in.py
```diff
@@ -33,7 +33,7 @@ def test_configset_from_ConfigSets(tmp_path):
     assert len([at for at in c_ats]) == 4
 
 
-@pytest.mark.xfail()
+@pytest.mark.xfail(reason="should fail creating ConfigSet_in from mix of ConfigSet_ins using input files and input configs")
```
gelzinyte, line 36: Why are you using xfail? My understanding is that `xfail` is for tests that fail because the code is incomplete or yet to be fixed and the cases the code is expected to fail should be handled via `with pytest.raises(RuntimeError): ...`. It seems to me that this is the latter case?

gelzinyte:
> I guess I understand - the iterator is just supposed to skip groups that contain None. The real problem is that the earlier minims modify the Atoms structures in memory, so they both converge even in the 3rd call. Fixed now. Take a look at ny next push in this PR.

Exactly, thanks for clearing up the bug.

tests/test_configset_in.py
bernstei: Yes - I think I misunderstood xfail.  I'll update that.  We should also check to make sure xfail isn't being used incorrectly in other places, I suppose.

bernstei: OK, fixed now.  The only other use of xfail is in the QE test, and that's the way it's intended.

bernstei:
BTW, I created issue #22 to remind us to deal with the QE test. 

@gelzinyte all the conversations are resolved, so post a note here if you think it's ready to merge, or if there's anything else to check.

gelzinyte:
> One thing we could do is require that there be a checked in file with the results of a complete test, and that there are no unexpected failures. It would mean that for any PR to be merged someone who can do such as test will have to grab it, run the test, and push an update version of that file. Is that only me, or is anyone else available? Ideally this would be a single machine where you could run the tasks locally as well as submitted to a queuing system (even better would be a queuing system on a remote machine, which needs to be accessed with passwordless ssh).

I think that sounds reasonable. Since we're reviewing the (major) pull requests, it could be up to the reviewer to check either that the file is included or run and update the full tests results. I am definitely available.

"there be a checked in file" - does that mean git-tracking a pytest results' file or is there an alternative mechanism to check in a file? 

All the tests have passed for on my end except I haven't checked those for vasp (I don't think we have a license) and test_cli_rss (it's been running for around two hours - is that ok or too slow?). Do merge if that sounds good to you.

bernstei:
> I think that sounds reasonable. Since we're reviewing the (major) pull requests, it could be up to the reviewer to check either that the file is included or run and update the full tests results. I am definitely available.
> 
> "there be a checked in file" - does that mean git-tracking a pytest results' file or is there an alternative mechanism to check in a file?

Yes - I came up with a somewhat elaborate scheme: there would be a script that's run manually after the last real commit that runs the tests, confirms that they are as expected (I have this part already, maybe not pushed), and commits some output file indicating the revision that is currently checked out and the success of the tests.  Then the CI testing would need to ensure that the file exists, and that the last commit is exactly one after the one indicated in the output file, and the only change between the two is the updated test status file.  I think that would be a straightforward thing to do in the github actions yml, if it's possible to get git history from the github action environment.  

That would ensure that the latest code version has been tested and passes.  As long as we allow PRs to be committed without requiring the tests be successful, we could then even choose to ignore that status if we wanted.  It would lead to some extra commits, with the testing results file that the CI would need to look at. 

> 
> All the tests have passed for on my end except I haven't checked those for vasp (I don't think we have a license) and test_cli_rss (it's been running for around two hours - is that ok or too slow?). Do merge if that sounds good to you.

Don't worry about VASP, but it would mean that I'd be the only person who could run the complete test (and if we ever had mandatory CASTEP tests, no one could run a complete test until I got myself a CASTEP installation, which in principle is possible, but I don't have access to right now).

As for the rss cli test, I think on my machine it usually takes more like 45 minutes. If you don't mind leaving it running overnight, I'd like to know what happens.

gelzinyte:
Ok, the rss cli test now successfully finished - it took 43 min submitted to a node. 

Ah yes there's `complete_pytest`. The scheme sounds good, but I think the test-output-tracking fille will have to change. New tests will get added and not everyone will be able to run all the tests (e.g. me and vasp), so automatically checking it might not be at all straightforward. I guess one low-tech solution could be to keep track of the test-output-tracking file and manually check for any differences and only approve the PR if there are none or if they are reasonable (should be trivial via `Files changed` Github action) . Also, probably thinking way ahead, I don't think it makes sense to make someone new contributing something smallish test everything and install programs they'd never use, for example airss or dft codes. ASE must be in a similar situation, do you know how they're dealing with their PRs & tests?

bernstei:
> Ok, the rss cli test now successfully finished - it took 43 min submitted to a node.

Great.

> Ah yes there's `complete_pytest`. The scheme sounds good, but I think the test-output-tracking fille will have to change. New tests will get added and not everyone will be able to run all the tests (e.g. me and vasp), so automatically checking it might not be at all straightforward. I guess one low-tech solution could be to keep track of the test-output-tracking file and manually check for any differences and only approve the PR if there are none or if they are reasonable (should be trivial via `Files changed` Github action) . Also, probably thinking way ahead, I don't think it makes sense to make someone new contributing something smallish test everything and install programs they'd never use, for example airss or dft codes. ASE must be in a similar situation, do you know how they're dealing with their PRs & tests?

The automatic part would just be confirming (in the CI, similar to the way it runs pytest and confirms that it hasn't failed) that the PR contains the right file.  After any random user contributes a PR, the automated testing will never show "All checks have passed".  At best it'll show that the actual pytest runs passed, but there will be another test (checking the "complete_test.out" file), which will fail because the git versions won't look up to date enough. Someone who _can_ run all the tests (me, you?) will have to actually do it, to switch the automated testing result to green.

bernstei:
I'm going to merge this one, and then create a PR for the automated check of the manual complete test, so we can see how it behaves.

PR #21, opened 2022-01-14
WIP: Extend ace fitting code with more functionality

gelzinyte:
Main additions: 

* Explicitly wrote out what was in `rpi_basis` and `pair_basis` to allow for element-specific cuttoffs
* Option to define polynomial degrees for each (element-specific) correlation order
* Clear up some of the variables' names

Left to do - clear up some redundant/inefficient code and expand tests.


bernstei:
Just remove the WIP when you want me to look at it.

gelzinyte:
> Just remove the WIP when you want me to look at it.

Ok, will do. This currently seems to work (or was working with ACE0.8) based on small tests, but before merging I would like to use this in my fitting workflow, just to check that I haven't missed something. 

Then if that looks good, we could merge and separate ace_fit.jl part into a separate repository (where?) in a following PR?

gelzinyte:
This is irrelevant after #33 functionality that was needed here should be implemented in ACE1pack.

PR #23, opened 2022-01-20
Change from ACE to ACE1

bernstei:
Change loaded module in ace_fit script, change check in test, and provide default Project.toml if user hasn't specified one in general or in a WFL specific env var

### DIFF
--- a/tests/test_ace_fitting.py
+++ b/tests/test_ace_fitting.py
@@ -13,7 +13,7 @@ import ase.io
 from wfl.configset import ConfigSet_in
 from wfl.fit.ace import fit, dict_to_ace_fit_string
 
-have_julia_with_modules = os.system("julia -e 'using ACE, IPFitting'") == 0
+have_julia_with_modules = os.system("julia -e 'using ACE1, IPFitting'") == 0
 
 def test_dict_to_ace_fit_string():
     # atomsfile: list of strings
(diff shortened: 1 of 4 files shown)


gelzinyte:
I would like to check that everything works before merging this and PR #21, so would like to wait for the ACE1 issues to be fixed first.

bernstei:
@gelzinyte turns out that JULIA_PROJECT can be handled entirely by the user, without any code changes.  I documented this in the ace fitting docstring, but since it also applies to generic calculator with `ACECalculator`, is there someplace better we can put it?

gelzinyte:
Probably documentation - at operations/calculators (from generic/ACECalculator point of view) and code_structure/environment variables (from env vars to be conscious of pow).

bernstei:
Reminder to check generic calculator with ACECalculator and `WFL_AUTOPARA_NPOOL`.  I'm having problems - it's either surprisingly slow, or just hanging.

gelzinyte:
The original description mentions a default `Project.toml`, but I don't see that among changed files, is that correct?

noambernstein:
On Wed, Feb 2, 2022 at 11:12 AM gelzinyte ***@***.***> wrote:

> The original description mentions a default Project.toml, but I don't see
> that among changed files, is that correct?
>

I decided against it. It gets messy especially with the remote jobs. I’d
prefer to try making the user manage their own. You can always copy your
Project.toml to the remote machine and set JULIA_PROJECT. We can change
later if we want.

gelzinyte:
I haven't yet checked for any issues with parallelising ACECalculator and I haven't been following the latest bug (segmentation faults) fixes, maybe that will have it covered. But otherwise yes, #33 covers this PR. If you are still having problems you mentioned above maybe open an issue and close this?

bernstei:
OK.  I'm closing this one.  When this PR is merged and the ACE1pack changes merged as well, I'll update, run the test with ExPyRe, and we can fix any problems then.

PR #24, opened 2022-01-20
Per remote job header

bernstei:
New field,  `header_extra`, in `RemoteInfo` structure, which is passed to `ExPyRe.start()`. 

Note that this PR is currently based on the ACE1 patch, which will therefore need to be merged first (or something will have to be rebased).


bernstei:
@gelzinyte if you think this one is OK too, I can merge it and @WillBaldwin0 can use it once the local expyre and workflow installations are both updated.

gelzinyte:
Looks good!

PR #28, opened 2022-01-24
converting Path to string in __str__

gelzinyte:


wfl/configset.py
```diff
@@ -315,7 +315,7 @@ def __str__(self):
         if self.input_configs is not None and len(self.input_configs) > 0:
             s += _fmt("input configs #", [str(len(at_group)) for at_group in self.input_configs])
         if self.input_files is not None and len(self.input_files) > 0:
-            s += _fmt("input files", [f"{fn} @ {idx}" for fn, idx in self.input_files])
+            s += _fmt("input files", [f"{str(fn)} @ {idx}" for fn, idx in self.input_files])
```
bernstei, line 318: Not a big deal, but this change isn't needed - the `f"{}"` already converts to string.  The one below, where the list elements are passed to `_fmt` and will be interpreted as strings, is needed.  However,  I would actually have suggested doing to casting to string in `_fmt`.  If you don't want the default string conversion you can always do it explicitly before passing to `_fmt`.

PR #33, opened 2022-02-01
WIP: ACE1pack update

gelzinyte:
Plan outline: 

- `ACE1pack.fit_ace()` takes in the parameters' dictionary and fits the model
- `ace_fit.jl` is a very thin interface thin interface that calls `ACE1pack.fit_ace()` and does what ACE1pack (currently) cannot:
  * do a workflow-style dry run
  * set BLAS number of threads from `ACE_FIT_BLAS_THREADS`
- `wfl.fit.ace.fit()` takes in fitting configs and parameters' dict, writes them to disk and calls `ace_fit.jl`, in a remote job if needed
- `wfl.fit.ace.prepare()` sets up the fitting configs and parameters' dict in such a way that behaviour of this branch is consistent with before. 
 
This way users can set up the fit differently from `wfl.fit.ace.prepare()` and still call `wfl.fit.ace.fit()` (as a remote job) from workflow/python.

Reminders: 
- From ACE1 PR #23: 
  * Check `ACECalculator` x `WFL_AUTOPARA_NPOOL` (either slow or hanging)
  * Double-check Julia project management (e.g. default `Project.toml` & handling of `JULIA_PROJECT`)

edit: add description


bernstei:
How is this related to the other ACE1 PR, #23 ?  Is it supposed to replace it?

gelzinyte:
Yes, this should replace ACE1 PR #23 and the changes I need from PR #21. ACE1pack functions will replace the fitting bit in `ace_fit.jl`. `ace.py` will need to be extended to do some of the setting up that currently is also being done in `ace_fit.jl` (see the outline I added, any thoughts?). ACE1pack is based on ACE1 already, so I'll include those updates from #23.

bernstei:
> Yes, this should replace ACE1 PR #23 and the changes I need from PR #21. 

You want to just `git merge` or `git cherry-pick` from those, and close them, then?

> (see the outline I added, any thoughts?)

where is this outline?

bernstei:
I guess for the outline you just mean in the first comment?  If so, it looks good.

bernstei:
@WillBaldwin0 says that using Agnesi transform is important for getting sensible pair potentials when the atoms are large (e.g. Cs). If we haven't already switched to that as the default, we probably should.

cortner:
I’m pretty sure Cas has less positive experience with Agnesi

gelzinyte:
Yes, I don't think it would be sensible to rush to have Agnesi as a default. I think the defaults should be left to be handled in ACE1pack, anyway. Regardless, I think it would be a matter of a keyword in ace fitting parameters to specify the favourite transform.

bernstei:
On further discussion I agree, but if @WillBaldwin0 is right about the bad behavior of the current default, we should probably  change _something_ (although I agree that the change should perhaps be in `ACE1pack`).  Is there anything I can do to help make progress with this PR?

WillBaldwin0:
I think I will do some more thorough digging into the transforms this week. It seems odd that this week (when I've been using close to unregularized least squares) I have consistently seen very different dimer curves between the two transforms, and no change in errors, yet Cas sees the opposite in every way. I am also wondering about whether the normalisation and orthogonality between the ACE and the pair potential is causing some of the effect here. I sort of think that we don't really have good options for the defaults, because we can't yet predict what will happen if we change the transforms for an arbitrary system. This might actually mean that the poly transform without the `r0` parameter is a better choice for a default, because otherwise we have more parameters which may not behave in a transparent way.

gelzinyte:
This should work with `eg/dev` branch of ACE1pack.  @bernstei what do you think? The biggest thing I didn't copy over from the main branch are the defaults - I think they should be set or left as mandatory in ACE1pack. Sorry it's taken a while!

bernstei:
> This should work with `eg/dev` branch of ACE1pack. @bernstei what do you think? The biggest thing I didn't copy over from the main branch are the defaults - I think they should be set or left as mandatory in ACE1pack. Sorry it's taken a while!

From this description it sounds good, but I'll take a look at the code changes.

As for defaults, I'm happy to generally stick with the defaults of the underlying things (packages, programs, etc) that are called, unless we have a specific reason to override it, say if it really doesn't make sense for any of our use cases, or if the underlying package is slow to adopt some change we want.

wfl/fit/ace.py
```diff
@@ -65,30 +68,98 @@ def fit(fitting_configs, ACE_name, params, ref_property_prefix='REF_',
         contructor to be used to run fitting in separate queued job
     ACE_FIT_JULIA_THREADS: used to set JULIA_NUM_THREADS for ace_fit.jl, which will use julia multithreading (LSQ assembly)
     ACE_FIT_BLAS_THREADS: used by ace_fit.jl for number of threads to set for BLAS multithreading in ace_fit
+
     """
+    # why can't we accept no prop prefix?
```
bernstei, line 73: I suspect I just copied this from GAP fit or something.  The problem is that the ASE extxyz reader interprets plain `"energy"`, etc, as quantities to put in a `SinglePointCalculator`, rather than plain info fields. That's a hassle, because those are accessed in a qualitatively different way, and things like the database modification routines (that are used by gap-rss to make the per-config sigmas) need to handle it differently.  Also, it means that if you ever calculate something with these configurations (e.g. to compute a fitting error) the calculated values override the existing reference values.  Basically, it's to avoid messy ASE behavior.

tests/test_ace_fitting.py
```diff
@@ -86,25 +78,39 @@ def test_ace_fit(request, tmp_path, monkeypatch, run_dir='run_dir'):
...
+        ace_fit_params=params, 
+        ref_property_prefix='REF_',
+        run_dir=str(run_dir), 
+        skip_if_present=True)
     time_actual = time.time() - t0
     print('ACE', ACE)
 
     assert os.path.exists(os.path.join(tmp_path, run_dir, f'ACE.B_test.json'))
-    assert os.path.exists(os.path.join(tmp_path, run_dir, f'ACE.B_test.yace'))
+    # assert os.path.exists(os.path.join(tmp_path, run_dir, f'ACE.B_test.yace'))
```
bernstei, line 104: I'd like some sort of support for yace, but I'm happy to decide that there should be a JSON -> anything else converter in ACE1, rather than coding different format/suffix logic into the workflow wrapper.  Just making a note about it.

wfl/fit/ace.py
```diff
@@ -118,9 +189,8 @@ def _check_output_files(file_base, formats):
 
         xpr = ExPyRe(name=remote_info.job_name, pre_run_commands=remote_info.pre_cmds, post_run_commands=remote_info.post_cmds,
                       env_vars=remote_info.env_vars, input_files=input_files, output_files=output_files, function=fit,
-                      kwargs= {'fitting_configs': fitting_configs, 'ACE_name' : ACE_name, 'params': params,
-                               'ref_property_prefix': ref_property_prefix,
-                               'run_dir': str(run_dir), 'formats': formats, 'ace_fit_exec': ace_fit_exec,
+                      kwargs= {'fitting_configs': fitting_configs, 'ace_fit_params': ace_fit_params,
+                               'run_dir': str(run_dir), 'ace_fit_exec': ace_fit_exec,
```
bernstei, line 193: What happened to `REF_property_prefix`?

wfl/fit/ace.py
```diff
@@ -138,50 +208,61 @@ def _check_output_files(file_base, formats):
...
+    use_params["data"]["fname"] = fit_cfgs_fname
 
-    fitting_line = dict_to_ace_fit_string(use_params)
+    ace_fit_params_filename = ace_file_base.parent / (ace_file_base.name + "_fit_params.json")
+    print(ace_fit_params_filename)
+    with open(ace_fit_params_filename, "w") as f:
+        f.write(json.dumps(use_params))
 
-    cmd = f'{ace_fit_exec} {fitting_line} > {ace_file_base}.stdout 2> {ace_file_base}.stderr'
+    cmd = f"{ace_fit_exec} --fit-params {ace_fit_params_filename} "
```
bernstei, line 236: Is it worth it to pass a JSON on the command line, so we don't have to write a temporary file?

bernstei:
@gelzinyte I had a few comments.  Do you think it's worth it for me to try to switch to `eg/dev` and play with it right now, or should I wait for your changes to be merged into ACE1pack?

examples/ace_fit_params.json
```diff
@@ -0,0 +1,90 @@
+{
```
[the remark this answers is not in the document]
gelzinyte: It is redundant and will be irrelevant once the documentation for ACE1pack is updated. Meanwhile I thought I'd add both so it's convenient to copy and modify.

wfl/fit/ace.py
```diff
@@ -65,30 +68,98 @@ def fit(fitting_configs, ACE_name, params, ref_property_prefix='REF_',
...
+
+    # calculate virial from stress, since ASE uses stress but ace_fit.jl only knows about virial
+    _stress_to_virial(fitting_configs, ref_property_prefix)
+
+
+    isolated_atoms = [at for at in fitting_configs if len(at) == 1]
+    if len(isolated_atoms) > 0 and "e0" in ace_fit_params.keys():
+        raise RuntimeError("Got e0 both as an isolated atom and in ace_fit_params")
+    if "e0" not in ace_fit_params.keys():
+        ace_fit_params["e0"] = _prepare_e0(isolated_atoms, ref_property_prefix)
```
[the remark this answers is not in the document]
gelzinyte: done

tests/test_ace_fitting.py
gelzinyte: Would you usually need either .json or .yace or both? If the former, ACE1pack could be modified to save to the appropriate format based on the suffix, if the latter, I think a converter script is better than having an option or default to save both both.

wfl/fit/ace.py
```diff
@@ -65,30 +68,98 @@ def fit(fitting_configs, ACE_name, params, ref_property_prefix='REF_',
...
-                    _ = json.load(fin)
-                elif fmt == '.yace':
-                    warnings.warn('Cannot parse yace format (can contain dicts with list keys), so just checking its existence')
-                    # _ = yaml.safe_load(fin)
-                else:
-                    raise ValueError(f'Cannot parse ACE file with format {fmt}')
-
-    # prefix each format by '.' if not there yet
-    formats = [('' if fmt.startswith('.') else '.') + fmt for fmt in formats]
+    ace_file_base = run_dir / (Path(ace_fit_params["ACE_fname_stem"]).name )
```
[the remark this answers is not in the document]
gelzinyte: Concatenating would get a bit complicated if the full path is specified, no? Also, how or where else could we "forbid" using paths if not by taking just the name of whatever was given? 

Would you ever find relative paths + concatenation useful? If not, I would rather not over-engineer and only have `run_dir/ace_filename.json` type of things.

wfl/fit/ace.py
gelzinyte: That is managed in `prepare_params()`. `fit()` doesn't modify the params beyond adding `.xyz` to fit to and optionally a `run_dir`.

wfl/fit/ace.py
```diff
@@ -138,50 +208,61 @@ def _check_output_files(file_base, formats):
...
-    for at in fitting_configs:
-        if ref_property_prefix + 'stress' in at.info:
-            stress = at.info[ref_property_prefix + 'stress']
-            if len(stress.shape) == 2:
-                # 3x3
-                at.info[ref_property_prefix + 'virial'] = list((-stress * at.get_volume()).ravel())
-            else:
-                # Voigt 6-vector
-                at.info[ref_property_prefix + 'virial'] = list((-voigt_6_to_full_3x3_stress(stress) * at.get_volume()).ravel())
+    fit_cfgs_fname_given = None
```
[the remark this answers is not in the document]
gelzinyte: Workflow paths override whatever was in the ace parameters given to these functions. That only applies to `ACE_fname` and `*params["data"]["fname"]`. In both cases, a warning is issued if something is actually being overwritten. 

`*params["data"]["fname"]` is mandatory in ACE1pack, but not in Workflow, because `fit()` takes configs to fit to and writes them to file itself. So there's some handling that in case some of the fields aren't present in the dictionary.

wfl/fit/ace.py
gelzinyte: I would like to write to file, because that makes it more transparent and easier to debug. The fitting conifgs, `.stderr` and `.stdout` are written to file anyway, so I don't think a parameters' file adds any more clutter. Also saves us from escaping special command line characters and `ace_fit.jl` needing to parse that (maybe the last bit isn't too big of a problem, though).

wfl/fit/ace.py
```diff
@@ -228,61 +309,20 @@ def _check_output_files(file_base, formats):
 
     return ace_file_base
 
+def _stress_to_virial(fitting_configs, ref_property_prefix):
+    for at in fitting_configs:
+        if ref_property_prefix + 'stress' in at.info:
+            stress = at.info[ref_property_prefix + 'stress']
+            if len(stress.shape) == 2:
```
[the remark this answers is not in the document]
gelzinyte: ok, updated in the next commit

tests/test_ace_fitting.py
```diff
@@ -11,36 +11,13 @@
...
-    # weights: dict
-    params = {'atoms_filename': 'dummy.xyz', 'outfile_base': 'dummy',
-              '_repeat_key': True, 'key': [['E', 'REF_energy'], ['F', 'REF_forces']], 'solver': [ 'lsqr', [[0.1, 1.0e-6]] ],
-              'weights': {'default' : { 'E' : 1, 'F' : 1, 'V' : 1}} }
-    s = dict_to_ace_fit_string(params).strip()
-    print('s', s)
-    assert s == '--atoms_filename dummy.xyz --outfile_base dummy --key E REF_energy --key F REF_forces --solver lsqr \'[[0.1, 1e-06]]\' --weights \'{"default": {"E": 1, "F": 1, "V": 1}}\''
+from wfl.fit.ace import fit, prepare_params_and_fit
 
+have_julia_with_modules = os.system("julia -e  'using ACE1pack'") == 0
```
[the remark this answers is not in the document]
bernstei: I was just thinking that the test can get done inside the testing function. That way if the testing function is run remotely it'll just fail at that point. I'm not sure there needs to be a separate test for the availability of julia before the function that tests the actual fit call.

examples/ace_fit_params.json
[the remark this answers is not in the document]
bernstei: OK

wfl/fit/ace.py
bernstei: OK.

wfl/fit/ace.py
```diff
@@ -15,29 +18,29 @@
 from expyre import ExPyRe
 import wfl.scripts
 
-def fit(fitting_configs, ACE_name, params, ref_property_prefix='REF_',
-        skip_if_present=False, run_dir='.',
-        formats = ['json', 'yace'], ace_fit_exec=str((Path(wfl.scripts.__file__).parent / 'ace_fit.jl').resolve()), dry_run=False,
-        verbose=True, remote_info=None, wait_for_results=True):
-    """Runs ace_fit on a a set of fitting configs
+
+def prepare_params_and_fit(fitting_configs, ACE_fname_stem, ace_fit_params, ref_property_prefix='REF_',
```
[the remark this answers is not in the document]
bernstei: OK, but I don't see this split in the repo code.  Am I correct that it hasn't been pushed yet?

wfl/fit/ace.py
bernstei: My first instinct is to say that it belongs in the `fit()` function, but I'll take a look at the revised version and think about this.

wfl/fit/ace.py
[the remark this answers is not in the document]
bernstei: That's what I figured, but it's still hard to figure out what's going to happen, with the parameters dict and the explicit function argument both being possible ways it can be specified.  That's why I want a comment explaining it.

tests/test_ace_fitting.py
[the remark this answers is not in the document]
gelzinyte: I see. I have updated the code, but I'm yet to test if the remote pytest.skip behaves correctly (likely not...).

wfl/fit/ace.py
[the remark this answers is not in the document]
gelzinyte: have pushed it now!

wfl/fit/ace.py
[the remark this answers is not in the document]
gelzinyte: Of course you meant the code. Have separated that out as a function for clarity and made the note.

gelzinyte:
Thanks a lot for the feedback, Noam! 

That PR is waiting for comments and/or approval from Christoph. I expect the user-facing parameters' dictionary not to change much or at all and any comments would be much appreciated.

tests/test_ace_fitting.py
bernstei: Let's just support only JSON here - Christoph says a conversion script is nearly trivial.

wfl/fit/ace.py
```diff
@@ -15,29 +19,74 @@
...
     Parameters
     ----------
     fitting_configs: ConfigSet_in
         set of configurations to fit
-    ACE_name: str
-        name of ACE potential, saved into files with yace and json suffixes
-    params: dict
-        dict of keys to turn into command line for ace_fit. each key will be prepended by
-        '-' (single characer) or '--' (longer name)
+    ACE_fname: str
```
bernstei, line 31: note explicitly that it'll override the name set in the params dict?

wfl/fit/ace.py
```diff
@@ -138,49 +167,71 @@ def _check_output_files(file_base, formats):
...
 
-    # write to file since that's what script needs as input
-    fitting_configs_filename = str(run_dir / f'fitting_database.{ACE_name}.extxyz')
-    ase.io.write(fitting_configs_filename, fitting_configs)
 
-    use_params["atoms_filename"] = fitting_configs_filename
+def _write_fitting_configs(fitting_configs, use_params, ace_file_base):
+    """
+    Writes fitting configs to file and updates ace fitting parameters. 
+    Configurations and filename handled by Workflow owervrite any filename
```
bernstei, line 194: typo

wfl/fit/ace.py
```diff
@@ -138,49 +167,71 @@ def _check_output_files(file_base, formats):
...
+    specified in parameters. 
+    """
 
-    fitting_line = dict_to_ace_fit_string(use_params)
+    # configs need to be in memory so they can be modified with stress -> virial, and safest to
+    # have them as a list (rather than using ConfigSet_in.to_memory()) when passing to ase.io.write below
+    fitting_configs = list(fitting_configs)
```
bernstei, line 200: do we need to do this conversion here given that we do it elsewhere (`prepare_configs`)

wfl/fit/ace.py
bernstei: OK, I understand what it's doing.  I'm definitely not sold on the fact that now you need to call 3 separate routines to do a fit.  At least I'd like a function that calls all 3 (prep params, prep configs, fit). I'd actually argue that this function should be called `fit`, like the GAP fitting codes, and this one that expects already prepared params and configs should be called something more specific.

tests/test_ace_fitting.py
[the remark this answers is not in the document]
bernstei: If I'm right that pytest.skip uses exceptions, and if you can pickle exceptions, we can make it work.  I hadn't thought about it, but if instead of printing the exception we pickle it, and then the local function creates it by unpickling and raising the resulting object, it might just work.

wfl/fit/ace.py
gelzinyte: no, we don't

wfl/fit/ace.py
gelzinyte: okay, I thought there should be a function that doesn't modify the parameters beyond what's necessary. Renamed it to   `run_ace_fit`, following the GAP example. `fit()` in the next commit should behave equivalently to `fit()` currently in `main`.

tests/test_ace_fitting.py
gelzinyte: ok!

wfl/fit/ace.py
bernstei: I'm happy with that.

tests/test_ace_fitting.py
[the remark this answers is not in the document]
bernstei: Let's not worry about this now, and hope the pytest.skipif will work. I'll play with the exceptions in expyre later.

wfl/fit/ace.py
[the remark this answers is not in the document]
bernstei: Well, you could in principle create all the temporary things in run_dir, but write the final potential to the full path.  I agree that it's not worth it.   In GAP multistage I assume that what's passed is a GAP _name_, i.e. a string, not a path. I'm happy to assume that here for ACE, but then using `Path(name).name` doesn't make sense. If we want to be very careful we can confirm that it's a string (or convert it to a string), and check that it doesn't contain any slashes.

bernstei:
I'd like to to another overall readthrough, but except the filename business I think it's probably ready to go.

wfl/fit/ace.py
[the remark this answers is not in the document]
gelzinyte: Okay, like in `gap_multistage.fit` let's ask for `ACE_name` and all of this should get resolved, I think. Let me know how does the latest commit look.

wfl/fit/ace.py
```diff
@@ -138,49 +228,63 @@ def _check_output_files(file_base, formats):
...
+    specified in parameters. 
+    """
 
-    fitting_line = dict_to_ace_fit_string(use_params)
+    if "data" not in use_params.keys():
+        use_params["data"] = {}
 
-    cmd = f'{ace_fit_exec} {fitting_line} > {ace_file_base}.stdout 2> {ace_file_base}.stderr'
+    fit_cfgs_fname_given = None 
+    if "fname" in use_params["data"].keys():
```
bernstei, line 263: No need for the `.keys()` part - `in dict` is enough.  Also, maybe cleaner with `fit_cfgs_fname_given = use_params["data"].get("fname", None)`?

tests/test_ace_fitting.py
```diff
@@ -49,25 +23,39 @@ def test_ace_fit_dry_run(request, tmp_path, monkeypatch, run_dir='run_dir'):
     (tmp_path / run_dir).mkdir()
 
     fit_config_file = os.path.join(os.path.dirname(request.fspath), 'assets', 'B_DFT_data.xyz')
```
bernstei, line 25: Since we're using `pathlib.Path` elsewhere, can we get rid of all the `os.path` syntax?

bernstei:
For some reason I'm having a very hard time following the whole name/path/run_dir logic when I look at the diffs in github.  I'm just going to check it out and clean it up how I like.  @gelzinyte Are you OK with me pushing that, and we can undo my changes if you don't like them.

wfl/fit/ace.py
bernstei: Doing this as part of my final (?) round of changes.

bernstei:
Thanks for merging.  Do you want to test it yourself, to make sure that it still works with the proposed `ACE1pack` changes?

gelzinyte:
Go ahead and merge this if you have no other comments.

bernstei:
I'm happy with it, but I'll wait for the tests to pass before I merge.  If they don't, we can discuss what if anything needs to be done.

gelzinyte:
That has been merged! So from a fresh julia environment, `add https://github.com/ACEsuit/ACE1pack.jl.git` will install everything you need

PR #34, opened 2022-02-15
clean up filename/path logic in ace1pack-update

bernstei:
Small changes to `wfl/fit/ace.py`, mainly to filename variables and logic.
Also remove `.keys()` from many instances of `if X in dict.keys()`.

### DIFF
--- a/wfl/fit/ace.py
+++ b/wfl/fit/ace.py
@@ -71,25 +71,30 @@ def fit(fitting_configs, ACE_name, ace_fit_params, ref_property_prefix='REF_',
     ACE_FIT_BLAS_THREADS: used by ace_fit.jl for number of threads to set for BLAS multithreading in ace_fit
     """
 
-    ace_fit_params = prepare_params(fitting_configs, ace_fit_params, ref_property_prefix)
+    ace_fit_params = prepare_params(ACE_name, fitting_configs, ace_fit_params, run_dir, ref_property_prefix)
     fitting_configs = prepare_configs(fitting_configs, ref_property_prefix)
 
-    return run_ace_fit(fitting_configs, ACE_name, ace_fit_params,
+    return run_ace_fit(fitting_configs, ace_fit_params,
                 skip_if_present=skip_if_present, run_dir=run_dir, ace_fit_exec=ace_fit_exec, dry_run=dry_run,
                 verbose=verbose, remote_info=remote_info, wait_for_results=wait_for_results)
 
 
-def prepare_params(fitting_configs, ace_fit_params, ref_property_prefix='REF_'):
+def prepare_params(ACE_name, fitting_configs, ace_fit_params, run_dir='.', ref_property_prefix='REF_'):
     """Prepare ace_fit parameters so they are compatible with the rest of workflow.
     Runs ace_fit on a a set of fitting configs
 
     Parameters
     ----------
+    ACE_name: str
+        name of ACE model, used as initial part of final potential JSON file, as well as other scratch files.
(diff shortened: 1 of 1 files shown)


wfl/fit/ace.py
```diff
@@ -30,17 +30,17 @@ def fit(fitting_configs, ACE_name, ace_fit_params, ref_property_prefix='REF_',
...
-        name of ACE potential (a .json). Overwrites any fileme given in the `params`. 
+        name of ACE potential (i.e. stem for resulting .json file). Overwrites any filename given in `ace_fit_params`.
     ace_fit_params: dict
-        parameters for ACE1pack. 
-        Any file names (ACE, fitting configs) already present will be updated, 
-        proprty keys to fit to will be prepended with `ref_property_prefix` and e0 
-        set up, if needed. 
+        parameters for ACE1pack.
+        Any file names (ACE, fitting configs) already present will be updated,
+        proprty keys to fit to will be prepended with `ref_property_prefix` and e0
```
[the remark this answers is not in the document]
bernstei: Fixed.  Let me know how you want to deal with these suggested changes.

wfl/fit/ace.py
```diff
@@ -69,47 +71,40 @@ def fit(fitting_configs, ACE_name, ace_fit_params, ref_property_prefix='REF_',
     ACE_FIT_BLAS_THREADS: used by ace_fit.jl for number of threads to set for BLAS multithreading in ace_fit
     """
 
-    ace_fit_params = prepare_params(fitting_configs, ACE_name, ace_fit_params, ref_property_prefix)
+    ace_fit_params = prepare_params(fitting_configs, ace_fit_params, ref_property_prefix)
```
gelzinyte, line 74: I think `ACE_name` is a workflow thing, so should be added in `prepare_params`, not `run_ace_fit`, which should take `ace_fit.jl`-ready parameters. I guess this is extra minor and idea of "`run_ace_fit` shouldn't modify the parameters" is violated anyway because we do need to modify them with fitting configs filename and run_dir, if applicable, but what I suggested would be consistent with `run_gap_fit()` and `gap_multistage.fit()`

bernstei: I did it this way because I didn't like the idea of setting `ACE_fname` to something we know might be wrong in `prepare_params`, and we can't set it to the right thing because it needs to include `run_dir`, but you don't know it until you get to `run_ace_fit`.  I guess the right thing in this framework is to pass `run_dir` to `prepare_params`, because the `ACE_fname` param depends on it.  Do you think that'd be better?

gelzinyte: Yes, let's do that!

bernstei: OK, I'll push an update (to this PR) soon.

bernstei: Actually, I just realized (as I was fixing the failed flake8 lint CI error) that because we only wrap `run_ace_fit` in `ExPyRe`, it's easier if the config file writing happens there.  If it happens before, we need to add the configs file to the `input_files`.  Or we could move the `ExPyRe` object and wrap the `fit()` call instead.

gelzinyte: Let's leave the fitting configs as they are then. ExPyRe should definitely be in `run_ace_fit` - I imagine someone preparing the configs & parameters themselves and just wanting a Python/ExPyRe interface to fitting the potential.

bernstei:
If you're happy with it and ready to merge into your branch, please do it.

gelzinyte:
Thanks for clearing all of this up! Would you like to merge this into #33?

PR #37, opened 2022-02-22
Fix scaling of energy_sigma, etc, when set by modify_database functions

### LINKED ISSUE #36: Bug: Potential issue when setting sigmas in gap_rss_set_config_sigmas_from_convex_hull.py
I have a gap-rss run started on 29th Nov 2021, with code version git https://github.com/gabor1/workflow/commit/968765772e20d9bab4e8ae21633a7a90bdb08e96-dirty.
This is related to the sigma-setting method in wfl/fit/modify_database/gap_rss_set_config_sigmas_from_convex_hull.py, but I noticed this method didn't change since, so the issues should still stand.

In my gap-rss run, I noticed in the training set for the final iteration that there are structures with energy_sigma = 0.001 (and 18 atoms say). 

However, when I recently tried to fit a GAP to my own structures, I found reasons to believe that the energy_sigma and virial_sigma values when set per config, are in fact not per atom, but are taken as a total for the atoms in that structure. So for example the energy_sigma = 0.001 for 18 atoms above corresponds to 0.0000(5) per atom. 

 In the attached picture I show a case where I set the sigmas to around 1meV/atom for each of the structures, but the resulting errors/atom are orders of magnitude smaller; hence the suspicion.image 
 
 This is corroborated by the bad force component fits from the gap-rss (note if github theme is dark one cannot read the labels): 
image
And the table of errors, where the force data has much larger errors in ratio to their sigma, than energy or virials do.
Screenshot 2022-02-22 at 18 00 37

I spoke with Gabor and he concurs that when set per config, energy_sigma, virial_sigma and hessian_sigma should be scaled by the number of atoms. While force_sigma should not be scaled.

bernstei:
Scale programmatically set energy, virial, and hessian sigmas by `len(at)`, as per #36.

WARNING: Will completely break backward compatibility of existing runs, since ratios between energy/force/virial sigmas will now be qualitatively different.

Closes #36

### DIFF
--- a/wfl/fit/modify_database/gap_rss_set_config_sigmas_from_convex_hull.py
+++ b/wfl/fit/modify_database/gap_rss_set_config_sigmas_from_convex_hull.py
@@ -128,10 +128,14 @@ def modify(configs, overall_error_scale_factor=1.0, field_error_scale_factors=No
             sigma_E_1 = 0.001
             sigma_E_2 = 0.100
             sigma_set = piecewise_linear(dE,
-                                         [(0.1, [sigma_E_1, np.sqrt(sigma_E_1), 2.0 * np.sqrt(sigma_E_1),
-                                                 2.0 * np.sqrt(sigma_E_1)]),
-                                          (1.0, [sigma_E_2, np.sqrt(sigma_E_2), 2.0 * np.sqrt(sigma_E_2),
-                                                 2.0 * np.sqrt(sigma_E_2)])])
+                                         [(0.1, [len(at) * sigma_E_1,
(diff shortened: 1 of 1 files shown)


bernstei:
@gabor1 I don't see Vlad, who found the issue, as a possible reviewer, so I assigned you.  If you want to have someone else look at it, since you're not familiar with the code, let me know.  It's nearly trivial, but completely changes the sigma heuristics and won't be backward compatible.

bernstei:
Will probably be subsumed into another PR that's a rollup of various updates.

bernstei:
replaced by #59, which includes this as well as other changes.

PR #39, opened 2022-02-25
Make calculators more robust to arbitrary ASE exceptions

bernstei:
CASTEP, Espresso, and ORCA calculators ignore _all_ exceptions n underlying ASE Calculator call, not just specific list.

### DIFF
--- a/wfl/calculators/orca.py
+++ b/wfl/calculators/orca.py
@@ -172,8 +172,8 @@ def evaluate_op(atoms, base_rundir=None, dir_prefix="ORCA_",
         try:
             at.calc.calculate(at)
             calculation_succeeded = True
-        except CalculationFailed:
-            pass
+        except Exception as exc:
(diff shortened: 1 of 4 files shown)


wfl/calculators/orca.py
```diff
@@ -172,8 +173,11 @@ def evaluate_op(atoms, base_rundir=None, dir_prefix="ORCA_",
         try:
             at.calc.calculate(at)
             calculation_succeeded = True
-        except CalculationFailed:
-            pass
+            if 'DFT_FAILED_ORCA' in at.info:
```
gelzinyte, line 176: Are these be coming from previous failed attempts?

bernstei: Yes, just in case it was set in an earlier attempt.

bernstei:
Abandoning - just part of #40 now

PR #42, opened 2022-02-28
Fix bug in remote job ACE fit

bernstei:
Current code converts fitting configs to list in `prepare_configs()`, but `run_ace_fit()` still tries to call the `to_memory()` method when preparing for a remote job.  PR removes the call to `to_memory()`, and documents that `run_ace_fit()` expects a _list_ of Atoms.

This would probably have been caught if we routinely ran the remote versions of the tests.

### DIFF
--- a/wfl/scripts/ace_fit.jl
+++ b/wfl/scripts/ace_fit.jl
@@ -34,6 +34,8 @@ function save_dry_run_info(fit_params)
 end
 
 if haskey(ENV, "ACE_FIT_BLAS_THREADS")
+    using LinearAlgebra
+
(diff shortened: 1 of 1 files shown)


bernstei:
@gelzinyte do you want to glance before I merge?

bernstei:
Also going to fix a bug in the new `ace_fit.jl`, which I noticed at the same time.

bernstei:
Oops - looks like I accidentally let the main patch get included in another recent PR, so this one is nearly trivial.  I'm just going to merge it, because the change to `ace_fit.jl` really is needed, but we should keep an eye on ACE fitting, I guess.

PR #60, opened 2022-04-14
Geometry optimisation - keep constraints from atoms objects 

LarsSchaaf:
Constraints (ie FixedAtoms) that are part of the atoms object that is being optimised get overwritten when specifying the symmetry command. Now its possible to clear the ase.Atoms object's constraints as previously or keep them.


bernstei:
I tend to lean to making things simpler, so not having yet another flag to clear constraints.  Just always append the symmetrization constraint to any existing ones, and remove it at the end.  Can you implement that?

bernstei:
Good, except that `clear_constraints` is still an argument to the low level function, and it's just ignored. 

How hard would it be to add a test for this (i.e. do a minim with a constraint and symmetrization and confirm that the constraint is there at the output)?

LarsSchaaf:
Yes can do !

LarsSchaaf:
Had accidentally formatted the test script with `black` [reason for forced push to my own fork]. 

Let me know if the test is what you imagined.

bernstei:
Looks good, but do you understand why the tests aren't running?  Is it just because it's in your fork?

bernstei:
@LarsSchaaf can you create this branch in _this_ repo, so we can see the test results?

bernstei:
I'm happy to merge, but curious - why did you change the test? Was it failing when you compared actual constraint objects?

LarsSchaaf:
Yes exactly. I didn't really see a quick way to compare constraint objects. Happy to change it if there's a trivial way!


https://stackoverflow.com/questions/1227121/compare-object-instances-for-equality-by-their-attributes

bernstei:
Thanks.  I'll merge now.

PR #63, opened 2022-04-28
Refactor and update ORCA calculators

gelzinyte:
Main changes: 
* Rename ExtendedORCA -> ORCA and separate BasinHoppingORCA into a different submodule,
* Move directory handling into the calculator itself and out of `evaluate_op`,
* Optional post-processing step with a function for natural population analysis.

@stenczelt, @bernstei any comments? @bernstei, especially on how files are handled in `ORCA.calculate()`? Now `ORCA()` calculator can be parallelised with `wfl.calculators.generic` and no `evaluate` and `evaluate_op` are needed, see `tests/calculators/test_orca.py::test_orca_with_generic()` for an example. I would like to rewrite other DFT calculators in the same way, so that the only way to parallelise property evaluations is via the generic calculator.


bernstei:
I'll take a look, but in general not needing a special calculator would be nice.

wfl/calculators/generic.py
```diff
@@ -57,14 +57,15 @@ def run_op(atoms, calculator, properties=None, output_prefix='_auto_', verbose=F
             # don't have that as default
             at.calc.calculate(at, properties=properties, system_changes=all_changes)
             calculation_succeeded = True
+            if f'{output_prefix}calculation_failed' in at.info:
+                del at.info[f'{output_prefix}calculation_failed']
         except Exception as exc:
             warnings.warn(f'calculation failed with exception {exc}')
+            at.info[f'{output_prefix}calculation_failed'] = True
```
bernstei, line 64: I don't love this way of doing error passing. For one, the calling function doesn't necessarily know the name of the info key where the failed status will be stored, since the default `output_prefix` will end up as different things. In addition, the particular exception is only saved in stderr (via warnings), and it might be nice to have it saved.  I'm not sure what to do about the first issue - maybe there's no good alternative, and I guess we're already accepting that the results are saved in keys that aren't necessarily known by the calling routine (so maybe eliminate `_auto_`?).  What do you think about saving the exception (convert to string) is saved in the `{output_prefix}calculation_failed` key?  Also, should we have an optional argument (default False?) which can be used to instead of ignoring it, actually raise the exception?

wfl/calculators/orca/bh.py
```diff
@@ -1,68 +1,16 @@
 import os
```
bernstei, line 1: rename this file to "basinhopping.py"?

bernstei:
More generic idea, since you had to add `remote_info` here - do we need a data structure (class?) for all the autoparallelization arguments, e.g. `npool`, `remote_info`, etc, so if we ever want to add additional ones we don't have to do it in every routine?

wfl/calculators/generic.py
gelzinyte: If there's ever need for the calling function to know `output_prefix`, it should simply be specified? I never leave it to `_auto_` and would be for making `output_prefix` mandatory to specify. 

Here I have copied what's in all the calculators (e.g. line 108 of castep.py), and I see it as more of a convenience flag rather than a way to pass errors? I don't really see a problem with the exception being written to stderr/warnings and can't really imagine needing it associated with `Atoms`. Maybe in more complicated workflows, but for me it would be enough to spot that some calculations have failed, check the reason in the output, fix them and re-submit just the configs I need. I think having the exceptions in `info` would be a bit cumbersome and more often than not will have to then be deleted. 

I thought I've seen calculator exceptions being optionally raised, but maybe I'm mistaking that with `skip_failed` of `iterable_loop`. I've just added a `raise_calc_exceptions` flag for this.

wfl/calculators/orca/__init__.py
```diff
@@ -0,0 +1,425 @@
...
+        self.post_process = post_process 
+
+
+    def calculate(self, atoms=None, properties=["energy", "forces"], system_changes=all_changes):
+        """Does the calculation. Handles the working directories in addition to regular 
+        ASE calculation operations (writing input, executing, reading_results) """
+
+        Calculator.calculate(self, atoms, properties, system_changes)
+
+        if self.scratch_path is not None:
```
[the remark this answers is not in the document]
gelzinyte: The reason I didn't put it in `__init__` is because `calculate` may be called multiple times within the same calculator and I think it's more robust to do every one of them in a clean directory. For example, orca would read in the wavefunction file if it's present from before and would crash if the geometries did not match. One of the reasons for ase to re-call `calculate` is change in `Atoms` geometry.

Are you also suggesting to get rid of one of the attributes relating to directory management?

gelzinyte:
I like the idea of collecting `iterable_loop` into a single structure - I have opened issue #65. I think that would make code easier to decode too. 

Feel free to merge once you're happy with the changes.

tests/calculators/test_orca.py
```diff
@@ -1,89 +1,245 @@
...
+                      task='gradient')
+
+ref_output = dict(energy=-2933.086022884649,
+                  forces=np.array([[-1.00418967e+03, -3.42059590e-07, 1.02844648e-05],
+                                   [1.00418967e+03, 3.42059590e-07, -1.02844648e-05]]),
+                  dipole=np.array([1.44071, -0.0, -0.0])
+                  )
+
+
+def test_extended_orca_calculator():
```
bernstei, line 39: why is this called `_extended_`?

wfl/calculators/generic.py
bernstei: Let's merge this now, but think in the long term about wether there's a better way to do this.

wfl/calculators/utils.py
```diff
@@ -147,6 +147,17 @@ def save_results(atoms, properties, results_prefix=None):
     if 'energies' in properties:
         atoms_results['energies'] = atoms.get_potential_energies()
 
+    if "extra_results" in dir(atoms.calc):
```
bernstei, line 150: What happened to the test confirming that `results_prefix` is not None when extra results are present?

tests/calculators/test_orca.py
gelzinyte: forgot to change the name, now fixed.

wfl/pipeline/utils.py
```diff
@@ -44,7 +44,10 @@ def __init__(self, sys_name, job_name, resources, job_chunksize=-100, pre_cmds=[
 
         self.sys_name = sys_name
         self.job_name = job_name
-        self.resources = resources.copy()
+        if isinstance(resources, dict):
+            self.resources = resources.copy()
+        else:
+            self.resources = resources
```
bernstei, line 50: I don't think having different copy semantics for dicts and objects is a good idea.  Do we need to implement a copy method for the object?

gelzinyte: that sounds good, I'll open a new PR

wfl/calculators/utils.py
gelzinyte: I initially misunderstood the logic, added that back in.

wfl/pipeline/utils.py
gelzinyte: ha, of course there is `copy.deepcopy(object)`, which I think is all? `copy.copy(object)` is not enough, becaue of `Resources.partitions`.

bernstei: I see that you just handled this with a `deepcopy`.  I'm happy with that, so closing this conversation,

bernstei:
Happy with the new code, will merge once tests pass.

gelzinyte:
Thanks!

Issue #45, opened 2022-03-22
documentation fixes


bernstei:
docs:
 - move CUR from generation to selection section

docstrings:
 - make `flat_histogram` docstring better, since it was confusing to explain
 - add a docstring to `atoms_and_dimers.py::prepare`

bernstei:
closed by #64

Issue #48, opened 2022-03-22
replace `iterable_loop` with functools.partial-based decorator


bernstei:
Replace cumbersome 
```
def run(inputs, outputs, arg1, ...):
    return iterable_loop(inputs, output, op, arg1, ...)
```
with decorator (although not using `@` syntax) `iloop`, which based on ideas [here](https://pythonicthoughtssnippets.github.io/2020/08/09/PTS13-rethinking-python-decorators.html)

May be able to do

```
def run(inputs, outputs, arg1, ...):
     # some code
run = functools.partial(iloop, run)
```
if we can figure out how to do `run.__doc__ = run.__doc__.format(iloop_docstring_post=iloop_docstring_post)`

gelzinyte:
Just to make a note that current `iterable_loop` messages such as 
```
 Running <function evaluate_op at 0x145549c97160> with npool=16, chunksize=1
```
could be made slightly more helpful by displaying/having better names than `evaluate_op`, etc.

PR #64, opened 2022-04-29
Rename various modules and functions

### LINKED ISSUE #43: rename or move many objects/files
`ConfigSet_in` -> `ConfigSet`, `ConfigSet_out` -> `OutputSpec`
`pipeline/` -> `autoparallelize/`, `iterable_loop` -> `autoparallelize
All low level `*_op` -> `*_autopara_wrappable`
`generate_configs/` -> `generate/`
`select_configs/` -> `select/`

move `selection_space.py` -> `select/`
move `mpipool_support.py` -> `autoparallelize`
move `calc_descriptors.py` -> `descriptors/calc.py`

bernstei:
More sensible names and organization for some important (and less important) modules and functions

closes #43


bernstei:
@gabor1 @gelzinyte who should review this?

gelzinyte:
I'm looking through this

wfl/autoparallelize/pool.py
```diff
@@ -102,7 +102,7 @@ def do_in_pool(npool=None, chunksize=1, iterable=None, configset_out=None, op=No
 
     did_no_work = True
 
-    if isinstance(iterable, ConfigSet_in):
+    if isinstance(iterable, ConfigSet):
         items_inputs_generator = grouper(chunksize, ((item, iterable.get_current_input_file()) for item in iterable))
     else:
         items_inputs_generator = grouper(chunksize, ((item, None) for item in iterable))
```
gelzinyte, line 108: Line 112 often translates to 
```
 Running <function evaluate_op at 0x145549c97160> with npool=16, chunksize=1
```
which is not perfectly helpful, because most of the `op`'s names aren't very helpful. Could this also report the module of the function? Or alternatively, should we give better names to all of `*autopara_wrappable' functions?

bernstei: I agree that it's not helpful.  I wasn't thinking that they needed to have a sensible name, but because of this sort of thing they do. Maybe do that as part of the `autoparallelize` manual wrap to `iloop` that's in issue #48 ?

wfl/descriptors/calc.py
```diff
@@ -8,7 +8,7 @@
 except ModuleNotFoundError:
```
gelzinyte, line 8: Why not to rename this to `wfl/descriptors.py` or something similar? Especially because there is a `calc` function there already??

wfl/autoparallelize/pool.py
gelzinyte: Sounds good, I've just made a note there.

bernstei:
also closes #45

wfl/descriptors/calc.py
```diff
@@ -76,9 +76,9 @@ def calc(inputs, outputs, descs, key, local=False, normalize=True, composition_w
```
gelzinyte, line 76: Is it worth renaming `calc` to something more descriptive? I am thinking of using it in code - I would be inclined to rename it when importing (like in `devtools/scripts/pretty_pca.py`) so it doesn't clash with other `calc` methods (if there are any) and the code using it is more readable.

wfl/descriptors/calc.py
bernstei: Now I'm confused - this used to be `wfl/descriptors.py`  @gabor1 do you remember why we said we should move it into a subdirectory, given that there isn't anything else in `wfl/descriptors/`?

wfl/calculators/dft.py
```diff
@@ -61,26 +61,26 @@ def evaluate_dft(
...
         raise ValueError(f"Calculator name `{calculator_name}` not understood")
 
     # run the calculation in parallel
-    return iterable_loop(
+    return autoparallelize(
         iterable=inputs,
-        configset_out=outputs,
+        outputspec=outputs,
         op=op,
         base_rundir=base_rundir,
```
gelzinyte, line 85: Do you think `wdir` might be a bit more self-explanatory/common than `rundir`? I think `base` bit could be slightly more self-explanatory too, would `root` be any better? So `wdir_root`?

bernstei: I don't find `wdir` to be any better - maybe `workdir`?

wfl/descriptors/calc.py
bernstei: Can you do `from wfl import descriptors` and then use it as `descriptors.calc()`?  Anyway, let's wait until @gabor1 reminds me why we decided to move it into a subdirectory.

wfl/configset.py
```diff
@@ -324,7 +324,7 @@ def __str__(self):
         return s
 
 
-class ConfigSet_out:
+class OutputSpec:
     """Thin output layer for configurations into files, ABCD, or atomic configs
```
gelzinyte, line 329: I find the `all_or_none=True, force=True` combination for skipping operations that would overwrite the `OutputSpec` files a bit cryptic. Would it be worth adding a tag `skip_if_complete` (or similar) in addition or instead?

wfl/descriptors/calc.py
gelzinyte: Yes, that would work perfectly fine.

wfl/configset.py
bernstei: The problem is that the actual skipping functionality is in `autoparallelize`, not really a part of the output spec.   It would definitely be slightly more sensible if `force=True` wasn't necessary for the skipping.  

Anyway, I agree that putting those in is a bit obscure, but isn't that the default behavior now anyway, so users don't really have to worry about it?

bernstei: OK.  I stand by my earlier statement - the actual skipping functionality is in `autoparallelize()`, and it just requires an `OutputSpec` setting that makes it safe.  I also stand (mostly) by my initial claim that the `OutputSpec` arguments communicate what it's going to do, but it's just that it's not obvious that those behaviors are required to get safe skipping behavior.  The "mostly" is because I do agree that it'd be better if `force` didn't matter for the skipping.  I might be able to fix that, if we think it's worth it.

However, for what it's worth, you **do not** have to pass force=True.  If you pass `all_or_none=True` (which is actually the default now) and **no** `force`, it defaults to `force=True`, and the skipping works.

wfl/descriptors/calc.py
bernstei: renamed to `descriptors/quippy.py`

wfl/descriptors/calc.py
bernstei: It's now `descriptors.quippy.calc`. If that's OK, do you want to resolve this conversation?

wfl/calculators/dft.py
bernstei: Are you going to do that, or should I?

gabor1: I would prefer workdir

gabor1 (approved):
I scrolled through all the renamings and all looked good to me.

wfl/calculators/dft.py
gelzinyte: > Are you going to do that, or should I?
I'll do that in a bit

wfl/configset.py
gelzinyte: okay! I like that `all_or_none=True` (the default) sets the `force` as well. 

I think "all_or_none" came from slightly different reasoning than "skip if the file is present", but I think that's going to be the more frequent use./? What do you think about renaming "all_or_none" to reflect that and be slightly more self-explanatory?

wfl/generate/minim.py
```diff
@@ -7,7 +7,7 @@
 from ase.constraints import ExpCellFilter
```
gelzinyte, line 7: Do you think `optimize.py` (following ASE) would be ever so slightly clearer than `minim.py`?

wfl/generate/vib.py
```diff
@@ -14,8 +14,8 @@
 from scipy import stats
```
gelzinyte, line 14: let's rename this to `normal_modes.py`?

wfl/autoparallelize/base.py
```diff
@@ -86,14 +86,14 @@ def iloop(func, *args, def_npool=None, def_chunksize=1, iterable_arg=0, def_skip
     remote_info = kwargs.pop('remote_info', def_remote_info)
     label = kwargs.pop('label', def_label)
 
-    return iterable_loop(npool, chunksize, inputs, outputs, func, iterable_arg, skip_failed,
+    return autoparallelize(npool, chunksize, inputs, outputs, func, iterable_arg, skip_failed,
```
gelzinyte, line 89: Did we talk about looking for clearer names for `chunksize` and `job_chunksize`? 

not sure if any of these are any better for `chunksize`: `serial_chunk_size`/`serial_chunksize`/`num_serial`/`num_serial_items`/`num_serial_ops`/`num_operations_in_serial`

instead of `job_chunksize` maybe `items_per_job`?

wfl/fit/gap_simple.py
```diff
@@ -4,7 +4,7 @@
 from pathlib import Path
```
gelzinyte, line 4: how about making a `gap` submodule? Then 
* `gap_simple.py` -> `gap/gap.py` or just in `gap/__init__.py` 
* `gap_multistage.py` -> `gap/multistage.py`
* `glue_2b.py` -> `gap/spline_2b.py` or `gap/glue.py` or `gap/glue_2b.py`
* `gap_relocate.py` -> `gap/relocate.py`

wfl/configset.py
bernstei: If we can come up with a better name, great, but since `OutputSpec` is not the same as `autoparallelize`, and it's `autoparallelize` that does the skipping, I don't think the "skipping" part makes sense to be in the `OutputSpec` arguments.  The standard name for this behavior (an operation that's all or none) is an atomic operation, so a more technically correct (if not necessarily more clear) name for that argument could be `atomic_write`, but I'm not sure that really helps you.   The fundamental problem is that there's nothing inherent to the `OutputSpec` that makes it skip - it's only `autoparallelize` that uses certain `OutputSpec` _behavior_.

wfl/generate/minim.py
bernstei: For the sake of compatibility, I suppose that's fine.

wfl/autoparallelize/base.py
bernstei: Yes, we did, but I think that could also be a different PR. We need to finish the workflow presentation/discussion,

wfl/autoparallelize/utils.py
```diff
@@ -14,7 +14,7 @@ class RemoteInfo:
     resources: dict or Resources
```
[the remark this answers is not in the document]
bernstei: Yes, please.

wfl/fit/gap_simple.py
bernstei: Sounds good, although I don't like only gap simple in `__init__`.

wfl/configset.py
gelzinyte: okay, I don't have any better ideas, let's leave it as is.

wfl/fit/gap_simple.py
gelzinyte: you don't like putting things in `__init__.py` or would you like to have fewer/more things in there than just `gap_simple.py` (if so, what?)?

gelzinyte:
should I make the changes we have agreed to?

bernstei:
> should I make the changes we have agreed to?

If you can do it, that'd be great.

wfl/fit/gap_simple.py
bernstei: I don't like putting only gap_simple there.  I'd say put `gap_simple` in `simple.py`, and could go either way on having an empty `__init__` or having references to both simple and multistage.

gelzinyte:
I'm good with the changes once we resolve the remaining conversations, thanks!

wfl/generate/vib.py
bernstei: Sure.  I'll do that.

bernstei: Actually, I'm thinking this isn't worth doing here. There will be probably be changes from the deprecation issue, and we need to think about how this should coexist with the (phonopy-based?) version that's applicable to periodic systems.

wfl/fit/gap_simple.py
bernstei: I guess I'll do this.

bernstei:
I just realized that not everything in what is now called `remoteinfo.py` is actually about `RemoteInfo`  - I forgot about `grouper()`. Maybe we should go back to a more generic name, or split off just the definition of `RemoteInfo`?

wfl/fit/gap_simple.py
bernstei: I think this is done. @gelzinyte can you resolve this conversation if you think it's OK?

bernstei:
As far as I know everything should be done.  Hopefully the tests will pass in the CI (locally they seemed OK).

wfl/fit/gap_simple.py
gelzinyte: looks good!

bernstei:
OK - merged the orca test PR, and will merge once all tests pass.

bernstei:
I had to tweak some tests that run locally but on in github CI, but assuming it passes again, I'll merge.

PR #69, opened 2022-05-03
Fix encoding of dictionary-type gap_fit_params values for gap_fit commandline

### LINKED ISSUE #68: conversion from dict to gap command string in gap_fit
The docstring for `dict_to_gap_fit_string` claims that dicts are turned into `:` separated strings (e.g. what you'd need for "config_type_sigma").  However, it doesn't seem to actually do this, there's no evidence of code that does this, and the simple gap fit pytest doesn't actually test this functionality.

bernstei:
Fix behavior of `dict_to_gap_fit_string()` in gap_simple.py so that dicts actually become colon separated lists (functionality is actually in `wfl.utils.quip_cli_strings()`

closes #68


gelzinyte (approved):
Looks good, thanks for fixing this.

PR #75, opened 2022-05-19
Add testing configs to gap-rss reevaluate step

bernstei:
Mainly add support for re-evaluating testing configs in gap-rss reeval step.
Some cleanup to support 0 vacancies, interstitials, and/or surfaces in gap-rss MD defect steps (including changed testing of `input_configs` content in `ConfigSet.__init__()`).
Bug fix for `iter_specific` in `Params` for more deeply nested sections (maybe consider rethinking `Params` in general).


bernstei:
Still waiting on a complete pytest run, which cannot be done in github CI. I'll note when that's all passing. Would probably be a good idea to add tests for the new stuff, but given the upcoming refactor, I don't think I'll actually do it right now.

bernstei:
Tests all pass. I'm ready to merge.  @gelzinyte @gabor1 anyone want to review?

PR #95, opened 2022-06-27
Update outdated variable/dict key names in docs 

bernstei:
Update "n", ConfigSet_in, and some mentions of "tasks" that are now just "cores")


gelzinyte:
Do merge, thanks!

Issue #65, opened 2022-04-29
Structure for autoparallelisation arguments


gelzinyte:
Collect all arguments for `iterable_loop`, so these could be more easily passed to functions using `iterable_loop` and changed without modifying the signature of the calling function.

PR #108, opened 2022-06-28
Use iloop for wfl.calculators.generic, as an example of cleaner new syntax

bernstei:
Modify iloop for calling syntax that is compatible with sphinx docstring parsing and autoparallelization automatic remote_info selection.  Use this new syntax to define `wfl.calculators.generic.run`

Fix some docs issues, including sphinx syntax errors, and recreate `overview.*rst` which used to be `code_structure.*rst`


wfl/autoparallelize/base.py
```diff
@@ -9,45 +9,133 @@
...
+
+_iloop_docstring_params_pre = (
+"""inputs: iterable({iterable_type})
+    input quantities of type {iterable_type}
+outputs: OutputSpec or None
+    where to write output atomic configs, or None for no output (i.e. only side-effects)
+""")
+
+_iloop_docstring_params_post = (
+"""ITERABLE_LOOP_RELATED:
```
gelzinyte, line 50: Should we rename `iloop` & related? Especially here, because this will go in the docstrings?

wfl/autoparallelize/base.py
```diff
@@ -9,45 +9,133 @@
...
+
+    - num_python_subprocesses: int, default os.environ['WFL_NUM_PYTHON_SUBPROCESSES']
```
gelzinyte, line 52: Sometime go we talked about making a class (like `RemoteInfo`) to gather all of the autoparallelization arguments in one place (issue #65), so it's clearer what arguments go with the function parallelised and and we can modify them without modifying the signature of the auto-parallelised function. Is this a good place to add that? Not sure how much work that'd be; I can also add this to a subsequent PR.

wfl/autoparallelize/base.py
```diff
@@ -68,33 +156,46 @@ def iloop(func, *args, def_num_python_subprocesses=None, def_num_inputs_per_pyth
...
+    # copy kwargs and args so they can be modified for call to autoparallelize
+    kwargs = kwargs.copy()
+    args = list(args)
+    if 'inputs' in kwargs:
+        # inputs is keyword, outputs must be too, any positional args to func are unchanged
+        inputs = kwargs.pop('inputs')
+        outputs = kwargs.pop('outputs')
+    else:
+        # inputs is positional, remove it from args to func
+        inputs = args.pop(0)
```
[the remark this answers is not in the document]
bernstei: I don't think so - my intent is that if `inputs` is passed to the _wrapped_ function as a  positional argument, it's always the first. What `iterable_arg` controls is where it's inserted into the arg list of the _low level op_.

I've been assuming that we (the developers) have perfect control over the wrapped function, but not necessarily the low level function. However, In every real case so far we're wrapping low level functions that we also wrote, so we can make sure they also take the iterable as the first argument. The idea of `iterable_arg` was to make it possible to wrap more arbitrary functions, which might take the input iterable in a different place.   Is that complexity worth it? Should we just get rid of `iterable_arg` and require that the low level functions take the iterable as their first argument?  The only downside is that you couldn't necessarily wrap arbitrary functions, if they happened to expect the iterable in a different position.

wfl/autoparallelize/base.py
bernstei: Good point. I'll do that, and resolve the conversation.

wfl/autoparallelize/base.py
bernstei: Now that it's an iterator, it's a bit cleaner than it used to be - at least you don't have to remember to add all the arguments when you define the wrapped function.  It would clean up the docstrings - right now every wrapped function will have the "ITERABLE_LOOP related" section. Maybe that means this is a good time to do this.  Let me try to code something up.  I'll let you know when there's something to review.

bernstei: What about calling it `AutoparaInfo`, by analogy to `RemoteInfo`?

wfl/autoparallelize/base.py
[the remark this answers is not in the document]
gelzinyte: >  I don't think so - my intent is that if inputs is passed to the wrapped function as a positional argument, it's always the first. What iterable_arg controls is where it's inserted into the arg list of the low level op.

Ah okay, that makes sense. I think let's leave it as is.

wfl/autoparallelize/base.py
bernstei: done

wfl/autoparallelize/base.py
gelzinyte: Yup that sounds good, thanks!

bernstei: Somehow this ended up as a much harder software engineering task than I envisioned.  Almost ready now.

bernstei: Updated now. @gelzinyte let me know what you think - you can see how `wfl/calculators/generic.py::run()` does the wrapping, and how `tests/test_autoparallelize.py::test_pool_speedup()` passes info at script calling time.

gelzinyte: This looks good, thank you for the update. I think it will make adding new functions & using parallelization much more intuitive.

PR #121, opened 2022-06-29
Added the example.iterative_gap_fit directory

Zausinator:
Added  a simple Python example of an iterative fitting workflow. Only requires ase, wfl, and quippy to work. Located in wfl.examples.iterative_gap_fit


gelzinyte:
Thanks for this, looks good! In particular I like having a runnable script in the examples. I think it would be great to add a test that runs `batch_gap_fit.py` just to test that it doesn't error - how tricky would that be? Superficially, I'd add a `test_batch_gap_fit.py` that imports `batch_gap_fit` and runs `batch_gap_fit.main()`. For reference, there are mpi tests in `tests/mpi`.

Zausinator:
That sounds sensible, and in theory that should be quite straightforward, I will take a look in the coming days!

Zausinator:
Ok, I'm back from the DPG and ready to tackle this! I'm already running into my first problem, which is how do I correctly use the autoparallelize function? 
Ideally I would like the MDs or minimizations to be run in parallel, but for reasons unknown to me they're all performed in series. Is there more to parallelization than setting ```os.environ['WFL_NUM_PYTHON_SUBPROCESSES'] > "0"```?

bernstei:
No, that should do it.  Do you want to post, or add to this branch, the runnable script that's not working for you?

Zausinator:
It's the ```examples/iterative_gap_fit/batch_gap_fit.py``` script. So I see that some slight edits have to be made (i.e. that the ```generate.md.sample``` function requires the ```temperature``` kwarg to be a float, not int), but the parallelization seems to be up-to-date. Is it an issue that I get the following printed output: 
```Running wfl.calculators.generic.run_autopara_wrappable with num_python_subprocesses=4, num_inputs_per_python_subprocess=None```
Is the None part of this the problem?

**Comment by noambernstein** [comment]
Could be - I’ll take a look.

**Comment by simonwengert** [comment]
Maybe this helps. 

I also had a look at this and was able to make it work with doing the change below in the `sample()` function in `wfl/generate/md`
```
-    def_autopara_info={"initializer":initializer}
+    def_autopara_info = kwargs.pop('def_autopara_info', dict())
+    def_autopara_info.update({"initializer":initializer})
```

Afterwards, I was able to set python_subprocess arguments by passing `def_autopara_info = {'num_python_subprocesses': <val-1>, 'num_inputs_per_python_subprocess': <val-2>}` to the sample() function

**Comment by Zausinator** [comment]
@simonwengert This works! I guess the only real problem was that the ```num_inputs_per_python_subprocess``` doesn't default to 1 if it isn't explicitly set. I'll open a new issue!

**Comment by bernstei** [comment]
This is a particularly messy part of the code.  I'll take a look.

**Comment by bernstei** [comment]
@Zausinator You could either merge the #150 branch into this one, and see if it fixes the issue, or wait for that PR to be merged into main and then merge main into this PR.

**Comment by bernstei** [comment]
@Zausinator #150 has been merged, so please merge `main` again, and confirm that the default num inputs is fixed now.

**Comment by Zausinator** [comment]
Thanks! Will do later today!

**Comment by Zausinator** [comment]
Alright, I *think* I did everything correctly. @noambernstein, the parallelisation works! @gelzinyte, I added a very barebones test script, is that roughly how you wanted it?

File: docs/source/workflows.mlip_fitting.md [hunk 1]
```diff
@@ -1,2 +1,190 @@
...
+```
+The dictionary of located in ```multistage_gap_params.json``` contains all the necessary GAP hyperparameters for a [multistage fit](wfl.fit.gap.rst). As we are only investigating Cu, the list of unique atomic numbers only has one value, ```Zs=[29]```. To include SOAP heuristics, we add a dictionary of length scales specific to the Cu atom. For more information, we refer to the [universalSOAP package](https://github.com/libAtoms/universalSOAP). To generalize this part of the code for other systems, we suggest using the previously VASP-calculated dictionary located there. The training data is located in the xyz-file ```EMT_atoms.xyz```. This file can be any ase-readable type file, and includes one structure with only one Cu atom, no periodic boundary conditions, and the property ```config_type=isolated_atom```. The energy of this structure being our default Cu energy.
+
+```python
+### Initial GAP training
+fit_idx = 0
+gap_name = f'GAP_{fit_idx}'
+GAP_dir = 'GAP'
+if not os.path.isdir(GAP_dir):
+  os.mkdir(GAP_dir)
```
bernstei, line 89: If you switched to `pathlib.Path` for your file objects, you could have cleaner code like `GAP_dir.mkdir(exist_ok=True)`.  Not sure it's worth it, but I generally like `Path` over string filenames and the various `os.*` functions.

examples/iterative_gap_fit/batch_gap_fit.py
```diff
@@ -0,0 +1,422 @@
...
+    kwargs
+        Any kwarg that is in the wfl.calc_descriptor calc function
+    Returns
+    -------
+    None, the descriptors are going to be found in the out_file
+    """
+    in_config = ConfigSet(in_file)
+    out_config = OutputSpec(files = out_file)
+
+    try:
```
bernstei, line 181: yaml is technically a superset of json, and so can read json files.

tests/test_batch_gap_fit.py
```diff
@@ -0,0 +1,27 @@
+import os, shutil, sys
+
+import numpy as np
+import pytest
+
+
+def test_batch_gap_fit():
```
bernstei, line 7: Should this be protected by tests for `gap_fit` exitsting, like in `tests/test_gap_fitting_simple.py`

bernstei:
I put in a few little comments.  Mostly just style, so feel free to decide you don't want to deal with it, but I think protecting the pytests with a test for `gap_fit` is a good idea.

Zausinator:
Thanks for the suggestions, I included them. Plus I reduced the number of iterations in the pytests to 2. To have more iterations there seemed like overkill and took forever.

bernstei:
Looks good.  I'll merge if the tests pass.

PR #125, opened 2022-07-14
replace routine specific vars like WFL_GAP_SIMPLE_FIT_REMOTE_INFO, all use WFL_EXPYRE_INFO

bernstei:
Replace routine-specific env vars with `RemoteInfo` dicts with values in `WFL_EXPYRE_INFO`.  

Refactor logic to get `remote_info` from `_autoparallelize_ll()` into `autoparallelize/utils.py::get_remote_info()`, so it can be used by other functions, and call it from all routines that need a `RemoteInfo` `kwargs` dict, in particular for now GAP and ACE fitting functions.

### DIFF
--- a/.github/workflows_assets/config.json
+++ b/.github/workflows_assets/config.json
@@ -1,10 +1,10 @@
 { "systems": {
     "github": { "host": null,
         "scheduler": "slurm",
-        "header": ["#SBATCH --nodes={nnodes}",
-                   "#SBATCH --ntasks={ncores}",
-                   "#SBATCH --ntasks-per-node={ncores_per_node}"],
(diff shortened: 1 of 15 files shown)


bernstei:
I think this is ready to merge.  The majority of the substantive changes were refactoring the remote_info selection from passed python variables and environment variables into a standalone function, and having the other routines that used to use env vars other than `WFL_EXPYRE_INFO`, namely the GAP and ACE fits, using the new function. The only non-trivial change (non-trivial to debug, at least) has to do with how the multistage GAP fit mangles the `remote_info` object which it passes to the individual GAP simple fits.

I also fixed a couple of places in the CI configuration that were outdated with the latest ExPyRe changes.

gelzinyte:
The changes look good to me, thanks for the update!

PR #128, opened 2022-07-28
debug hanging quippy descriptor calc

bernstei:


bernstei:
debugging done - discovered you have to set OMP_NUM_THREADS=1 for descriptor calc to not hang in CI.

PR #129, opened 2022-07-28
use forkserver for autopara Pool

bernstei:
Using start method `"forkserver"` for `multiprocessing.Pool`'s context should behave better with OpenMP.


bernstei:
Abandon, and start again - probably need both forkserver (so OpenMP works) and also OMP_NUM_THREADS=1 in tests (so it's efficient).

PR #130, opened 2022-07-28
debug Pool forkserver PR

bernstei:


bernstei:
Abandon, and start again - probably need both forkserver (so OpenMP works) and also OMP_NUM_THREADS=1 in tests (so it's efficient).

PR #131, opened 2022-07-28
Switch to forkserver for autopara multiprocessing.pool.Pool context

bernstei:
Appears to be necessary so it works with OpenMP. See https://github.com/isl-org/Open3D/issues/1552


bernstei:
abandon - forkserver not working with remote jobs.

PR #133, opened 2022-08-02
bug fix in autoparainfo

gelzinyte:
@bernstei very quick bug fix - it used to be impossible to overwrite the AutoparaInfo defaults, set when parallelising a function. Please merge if this looks ok.

### DIFF
--- a/wfl/autoparallelize/autoparainfo.py
+++ b/wfl/autoparallelize/autoparainfo.py
@@ -58,8 +58,8 @@ class AutoparaInfo:
             if not hasattr(self, k):
                 setattr(self, k, def_kwargs.pop(k, AutoparaInfo._kwargs[k]))
             elif k in AutoparaInfo._kwargs:
-                # it is a valid default, but was already set in AutoparaInfo, so not 
-                # overwriting with defaults
+                # it is a valid default, but was already set in AutoparaInfo by the user,
+                #  so not overwriting with defaults
                 del def_kwargs[k] 
 
 
--- a/wfl/autoparallelize/autoparainfo.py
+++ b/wfl/autoparallelize/autoparainfo.py
@@ -57,9 +57,14 @@ class AutoparaInfo:
         for k in AutoparaInfo._kwargs:
             if not hasattr(self, k):
                 setattr(self, k, def_kwargs.pop(k, AutoparaInfo._kwargs[k]))
+            elif k in AutoparaInfo._kwargs:
+                # it is a valid default, but was already set in AutoparaInfo, so not 
+                # overwriting with defaults
+                del def_kwargs[k] 
+
 
         if len(def_kwargs) != 0:
-            raise ValueError("def_kwargs contained unknown keywords {list(def_kwargs.keys())}")
+            raise ValueError(f"def_kwargs contained unknown keywords {list(def_kwargs.keys())}")
 
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


wfl/autoparallelize/autoparainfo.py
```diff
@@ -57,9 +57,14 @@ def update_defaults(self, def_kwargs):
         for k in AutoparaInfo._kwargs:
             if not hasattr(self, k):
                 setattr(self, k, def_kwargs.pop(k, AutoparaInfo._kwargs[k]))
+            elif k in AutoparaInfo._kwargs:
```
bernstei, line 60: Isn't this if unnecessary? It's inside a loop of `k in AutoparaInfo._kwargs`?

bernstei (commented):
Let me think about the overall logic.

bernstei:
I've looked over my code, and I don't understand the purpose of this PR. The way I intended it to work is as follows:

- e.g. for the generic calculator, the user's workflow script calls the wrapping `generic.run()` function, passing an `AutoPara_info` object in the `autopara_info` kwarg.
- The `run()` function passes this object, as well as the `def_autopara_kwargs` dict (set by the person who wrapped the function) to `autoparallelize()`
- `autoparallelize()` extracts the `AutoPara_info` object into `autopara_info`, or constructs a default one if it does not exist
- `autoparallelize()` then calls `autopara_info.update_defaults()`
- `update_defaults()` loops over all possible `AutoPara_info._kwargs` keys, and _if the object doesn't have that attribute set_ it sets it from the `def_autopara_kwargs` dict passed to `autoparallelize`, and if that doesn't exist from the class-wide defaults in `AutoPara_info._kwargs`.

The critical thing is that `update_defaults` is supposed to copy values from the class-wide defaults or the defaults hard coded in the wrapping if and only if the user passed an `AutoPara_info` object that doesn't already have that attribute set.  I looked at the loop you modified, and don't see why that's not working correctly.

Can you give an example of what you want to do that doesn't work the way you expect?

wfl/autoparallelize/autoparainfo.py
gelzinyte: that's right, it should just be else in that case. Will update.

gelzinyte:
What if the user has changed one of the two defaults? In that case `if not hasattr(self, k):` doesn't get triggered, the parameter, even though correct, doesn't get deleted from def_kwargs and def_kwargs has something remaining and the ValueError gets raised. Try this: 

```
gap_calc = (Potential, [], {"param_filename":"gap.xml"})
gap_calc_autopara_info = AutoparaInfo(num_inputs_per_python_subprocess=20)

generic.run(
    inputs=inputs,
    outputs=outputs,
    calculator=gap_calc,
    properties=["energy", "forces"],
    output_prefix="gap_",
    autopara_info=gap_calc_autopara_info
    )
```

bernstei:
I hadn't thought about the interaction with checking for unused defaults.  Let me look over your changes again with that in mind.

wfl/autoparallelize/autoparainfo.py
bernstei: OK, now I understand.  Yes - if we change the `elif` to `else`, I think it'll be good to go.

gelzinyte:
ok thanks, I will merge once the tests pass

PR #135, opened 2022-08-11
updated ace_fit script

gelzinyte:
Moved `ace_fit.jl` to live with ACE1pack. 

The path to the new `ace_fit.jl` in the ACE1pack package location needs to be given as a function argument or an environment variable ("WFL_ACE_FIT_EXEC"). That isn't ideal, but it's not quite obvious how to do that better in Julia (see [issue in ACE1pack](https://github.com/ACEsuit/ACE1pack.jl/issues/48))


gelzinyte:
The tests pass for me locally for ACE1pack v0.0.3

bernstei:
> The tests pass for me locally for ACE1pack v0.0.3

Do you understand why they fail in the CI?  Does the new version of ace_fit try to use mpi by itself somehow?

bernstei:
Changes generally look good.  What do you think about changing our environment variables that end in `_EXEC` (this one, maybe also gap fit) to end with `_COMMAND`, by analogy with ASE calculator convention?

gelzinyte:
> Do you understand why they fail in the CI? Does the new version of ace_fit try to use mpi by itself somehow?

It shouldn't. It looked like it was a github issue (asking for more, 2, MPI processes/slots than there were available) - let's see if it persists with a rerun. 

> What do you think about changing our environment variables that end in _EXEC (this one, maybe also gap fit) to end with _COMMAND, by analogy with ASE calculator convention?

That's clearer, I'll update.

gelzinyte:
Looks like the test failure was a fluke. Do merge if you're happy with this.

bernstei:
> Looks like the test failure was a fluke. Do merge if you're happy with this.

The tests are still running for me, but if they pass, I'll merge.

bernstei:
Is this the time to make it possible to specify the GAP fit command with an env var, similarly to ACE? The only real difference is the weird situation that quippy puts `gap_fit` in the path, while the ACE installation does not (but it could, or at least the python interface could).

gelzinyte:
There's no ACE python interface installation - it's still Julia, but it reads JSON written by Python/whatever. 

I wouldn't have added `WFL_ACE_FIT_COMMAND` if there was a way to make an ace_fit executable. Is there a way? Currently one needs to call the script directly and I haven't found a way to specify the path for julia to look at so you can just call `julia ace_fit.jl -p params.yml` without giving the full path to `ace_fit.jl`. For GAP, rather than setting an environment variable, I would just extend PATH to have `gap_fit` in it? It's straightforward to add gap_fit environment variable if you think it's worth it.

bernstei:
If the julia script started with `#!/usr/bin/env julia` it would run if you made it executable. The only real difference between ACE and GAP, in their current form, is whether they place the fitting script someplace that's in your path.

gelzinyte:
Of course there's the shebang..

I've added `WFL_GAP_FIT_COMMAND`, but the logic is different from ace_fit, since it's easy to set a default to `gap_fit`. I'd be okay with this, but I can also make the logic consistent (either the env variable or function argument to be prioritised in both) if you think that'd be better.

bernstei (changes requested):
In general good, except I think we should redo the priority on argument vs. env var vs. default (I don't like the order for GAP), and consistent for both.  How about, explicit argument first, then env var, then default (or crash for ACE):

```
def ace_fit(.... ace_fit_command=None, ...):
    .
    .
    if ace_fit_command is None:
        if "WFL_ACE_FIT_COMMAND" in os.environ:
            ace_fit_command = os.environ["WFL_ACE_FIT_COMMAND"]
        else:
            raise ValueError(...)
```
and same for GAP, except that instead of the `raise ValueError` we do `gap_fit_command = "gap_fit"`.

gelzinyte:
done.

PR #137, opened 2022-08-14
Update all parallelized functions to consistently use the new autoparallelize

gelzinyte:
Rewrite the parallelized functions that have been working with `_autoparallelize_ll` to instead use `autoparallelize`, following `generic.py` example.

### DIFF
--- a/tests/test_autoparallelize.py
+++ b/tests/test_autoparallelize.py
@@ -13,7 +13,7 @@ from wfl.autoparallelize.autoparainfo import AutoparaInfo
 
 
 def test_empty_iterator(tmp_path):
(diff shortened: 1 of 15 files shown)


bernstei:
Everything here looks good, and the tests pass, so we could probably merge.  I haven't cloned it locally so I could search, so it's not easy for me to tell - is this _every_ autoparallelized function, or just a start?

bernstei:
We should remove references to `__autoparallelize_ll` from `wfl/autoparallelize/__init__.py` - that should break everything that still tries to use it directly instead of the new `autoparallelize()`.

gelzinyte:
it's all autoparallelized function except from `dft.py` and basin hopping orca calculator. I thought I'd update CASTEP and VASP for the former soon anyway and the latter is slightly broken and I think was planned to be updated (@stenczelt was that the plan? )

PR #138, opened 2022-08-14
Process md on the fly

gelzinyte:
1. Added an optional `md.py::sample` argument (a function) that at the end of the MD sub-selects configurations from the trajectory. Since this is done at the "python subprocess" level the irrelevant configs aren't ever saved to an `OutputSpec`. 
2. Added an optional argument (a function) that evaluates the the `Atoms` of an MD snapshot every time it is recorded. The function takes in an `Atoms` object and returns a `bool`. If the function evaluates to `False` for N snapshots in the row, a `RuntimeError` is raised and MD is stopped. Useful for not waisting resources if the fitted potential is unstable, for example. 

Based off PR #137


bernstei:
> 1. Added an optional `md.py::sample` argument (a function) that at the end of the MD sub-selects configurations from the trajectory. Since this is done at the "python subprocess" level the irrelevant configs aren't ever saved to an `OutputSpec`.

looks good, but I'd prefer a more explicit name, say `sample_traj_func`. Is it possible to do what you need with a function that runs "interactively", i.e. during the MD run, rather than after the fact?  If so, that'd prevent the need to even save the unneeded configs at all, even temporarily before filtering into the final ConfigSet.

> 2. Added an optional argument (a function) that evaluates the the `Atoms` of an MD snapshot every time it is recorded. The function takes in an `Atoms` object and returns a `bool`. If the function evaluates to `False` for N snapshots in the row, a `RuntimeError` is raised and MD is stopped. Useful for not waisting resources if the fitted potential is unstable, for example.

Again, I'm thinking about a more general way to do that.  What about an `abort_func()` that returns a bool, but keeps track internally of how many times it's failed. For example, if you want to switch to aborting if you fail 6/8, you'd have to code that up specially in md.py, but if md.py just called the function, the function itself could have that cleverness in it.  I suppose you could just make the default number of Falses in a row to be 1, and then have all the cleverness in the function.

I'll look over the specific changes now.

tests/test_md.py
```diff
@@ -10,6 +10,8 @@
 
 from wfl.generate import md
 from wfl.configset import ConfigSet, OutputSpec
+from wfl.generate.md import select_every_10_fs_for_tests
```
bernstei, line 13: If these functions are for tests only, do they really belong in `wfl.generate.md`, or should they be moved into this file?

gelzinyte: the functions are only for tests, but I could not pickle them if they were defined in the same file. Any ideas how to get around that? If we can't work out how to define and pickle functions in the same file, I'll just move them to a file in the `tests` directory.

bernstei: Let me play with it a bit, and see if I can figure out how to do it cleanly.

bernstei: Your comment in the code suggests that you tried to define it as a fixture, and that failed. I think that's only true for `@` decorator wrapped functions (and that's why the autoparallelization wrapper is written the way it is).  But you can pickle a simple function that's defined in the same file.  The following patch works for me:
```
diff --git a/tests/test_md.py b/tests/test_md.py
index c2fa469..a647c4a 100644
--- a/tests/test_md.py
+++ b/tests/test_md.py
@@ -10,9 +10,11 @@ from ase.calculators.emt import EMT
 
 from wfl.generate import md
 from wfl.configset import ConfigSet, OutputSpec
-from wfl.generate.md import select_every_10_fs_for_tests
+# from wfl.generate.md import select_every_10_fs_for_tests
 from wfl.generate.md import check_validity_for_tests
 
+def select_every_10_fs_for_tests(traj):
+    return [at for at in traj if at.info["MD_time_fs"] % 10 == 0]
 
 @pytest.fixture
 def cu_slab():
diff --git a/wfl/generate/md.py b/wfl/generate/md.py
index 3487ea9..551b73e 100644
--- a/wfl/generate/md.py
+++ b/wfl/generate/md.py
@@ -258,12 +258,12 @@ def sample(*args, **kwargs):
 sample.__doc__ = autoparallelize_docstring(sample_autopara_wrappable.__doc__, "Atoms")
 
 
-# for tests, can't pickle if defined as a fixture
-def select_every_10_fs_for_tests(traj):
-    return [at for at in traj if at.info["MD_time_fs"] % 10 == 0]
+# # for tests, can't pickle if defined as a fixture
+# def select_every_10_fs_for_tests(traj):
+#     return [at for at in traj if at.info["MD_time_fs"] % 10 == 0]
 
 
 def check_validity_for_tests(at):
     if "5" in str(at.info["MD_time_fs"]):
         return False
-    return True
\ No newline at end of file
+    return True
```

gelzinyte:
> What about an abort_func() that returns a bool, but keeps track internally of how many times it's failed. For example, if you want to switch to aborting if you fail 6/8, you'd have to code that up specially in md.py, but if md.py just called the function, the function itself could have that cleverness in it. 

I am not entirely sure how that'd work. Wouldn't the propagator-attached function call a new instance of `abort_func()`, so global variables are of no use? How to track the number of failures otherwise? 

One slightly clumsy solution is for `abort_func()` to take `history` - a list of booleans from previous invocations and return `updated_history` of length `len(history) + 1`. Then within `abort_func()` you could implement some kind of logic based on `history` and raise an error from within. 

I would much prefer fewer restrictions on the `abort_func()` (i.e. it just have to take in an `Atoms` and return a boolean), because then it's easier to buy into the whole framework. 

I guess we could support both modes -
a) function takes one parameter (`Atoms`) and returns a boolean - md.py then raises an error if the boolean is False. 
b) function takes two arguments (`Atoms`, `history`) and returns list of booleans - md.py then leaves it for `abort_func()` to raise the error. 

What do you think?

bernstei:
The object oriented solution would be do define a class, with objects that save whatever internal state (like history) they need, and a method that returns a bool abort status, and pass those objects, rather than a function to the MD function.  That should work, including pickling.

What I had in mind for my original suggestion is something like

```
def workflow_step():

    history = []

    def abort_func(at):
        .
        .
        ok_status_of_at = .....
        history.append(ok_status_of_at)
        .
        .
        ok_status_of_history = .... 
        return ok_status_of_history

    sample(ConfigSet, OutputSpec,  ..... abort_func)
```
each call to abort_func will be using the same history variable, from the `workflow_step` function. I'm _not_ sure whether this will work with pickling, though.

I _really_ don't like hardwiring things like "OK N steps in a row" inside the md sample function. And since closures (the 2nd approach) may not even work, and are probably less familiar, I vote for the OO idea, which would be something like
```
class MDFailCheck:
    def __init__(self, n_steps):
        self.history = []
        self.n_steps = n_steps

    def atoms_ok(self, at):
        self.history.append(<some calculation on at>)

    def abort_status(at):
        self.atoms_ok(at)
        return <some calculation using self.history and self.n_steps>


sample( .... MDFail(10))
```
and `sample` would have to do something like
```
def sample(...    abort_check=None, ...):
     if abort_check is not None:
        if abort_check.abort_status(at):
            <abort>
```

We could even make `md_utils.py` file in `wfl/sample/` with some common use cases, or a user could define their own (we could also define an interface class, which actual useful classes can be derived from).

gelzinyte:
> I really don't like hardwiring things like "OK N steps in a row" inside the md sample function. 

I agree and thanks for the OO suggestion. I'll add an interface class and a simple example - for example, a check for any two atoms being closer than 0.5Å (alternative suggestions are very welcome).

tests/test_md.py
gelzinyte: I thought I have tried this, but apparently not! Updated and seems to work.

gelzinyte:
ok, now there is `wfl/generate/md/abort.py` with the base and example derived classes and a test. Let me know what you think.

tests/test_md.py
```diff
@@ -7,11 +7,21 @@
...
 from ase.calculators.emt import EMT
+from wfl.autoparallelize import autoparainfo
 
 from wfl.generate import md
 from wfl.configset import ConfigSet, OutputSpec
+from wfl.generate.md.abort import AbortOnCollision
 
 
+def select_every_10_fs_for_tests(traj):
+    return [at for at in traj if at.info["MD_time_fs"] % 10 == 0]
```
bernstei, line 18: I think `MD_time_fs` is a float, so you probably shouldn't use `% 10 == 0` with it. Not that I can think of a safe way to do this that isn't pretty complicated (save the time of last call, do some rounding, etc).

tests/test_md.py
```diff
@@ -109,3 +119,36 @@ def test_NVT_complex_ramp(cu_slab):
...
+    atoms_traj = md.sample(inputs, outputs, calculator=calc, steps=300, dt=1.0,
+                           temperature = 500.0, traj_subsampling_fun=select_every_10_fs_for_tests)
+
+    atoms_traj = list(atoms_traj)
+    atoms_final = atoms_traj[-1]
+
+    assert len(atoms_traj) == 31 
+
+
+def test_md_abortion_function(cu_slab):
```
bernstei, line 140: `abort`, not `abortion`

wfl/generate/md/__init__.py
```diff
@@ -206,9 +218,13 @@ def process_step(interval):
             at.info['MD_time_fs'] = cur_step * dt
             traj.append(at_copy_save_results(at, results_prefix=results_prefix))
 
-        # save config_type
-        for at in traj:
-            config_type_append(at, 'MD')
+        if traj_subsampling_fun is not None:
```
bernstei, line 221: Should this also be done on the fly, like the name of the PR suggests?

wfl/generate/md/abort.py
```diff
@@ -0,0 +1,41 @@
...
+import numpy as np
+
+class AbortBase(ABC):
+    """Base class used for checking and aborting MD simulation of `wfl.generate.md.sample()`"""
+    def __init__(self, n_failed_steps=1):
+        self.history = []
+        self.n_failed_steps = n_failed_steps
+
+    @abstractmethod
+    def check_if_atoms_ok(self, at):
```
bernstei, line 11: Why is this in the base class? Shouldn't it be up to each derived class to decide what infrastructure it needs to make the final "stop MD" decision?

wfl/generate/md/abort.py
```diff
@@ -0,0 +1,41 @@
...
+        """Method to check whether this trajectory step is acceptable. 
+           The method must append a boolean to `self.history`, 
+           based on which the MD gets aborted. 
+           All derived classes must implement this method."""
+        ...
+    
+    def should_stop_md(self, at):
```
bernstei, line 18: I think I'd prefer a shorter name (surprisingly?), maybe just `stop_md()`

wfl/generate/md/abort.py
```diff
@@ -0,0 +1,41 @@
...
+        """Returns a boolean based of which `wfl.generate.md.sample()`
+        aborts the simulation. Defaults to aborting if `n_failed_steps` in a row `check_if_atoms_ok()` 
+        are evaluated to False. Derrived classes may overwrite this."""
+        self.check_if_atoms_ok(at)
+        return np.all(np.array(self.history[-self.n_failed_steps:]) == False)
+
+
+class AbortOnCollision(AbortBase):
+    def __init__(self, clash_radius=0.5, n_failed_steps=3):
```
bernstei, line 27: name consistency: `collision_radius`

tests/test_md.py
gelzinyte: True, I didn't think about this too much because it's supposed to be just a simple test (so we know exactly how it behaves) and not a user-facing user-proof function. For what it's worth `10.5 % 10 = 0.5` so this particular example works ok...

wfl/generate/md/__init__.py
gelzinyte: I thought having it at the end would be more general - e.g. for CUR you'd need the whole trajectory, no?

wfl/generate/md/abort.py
gelzinyte: This is in the base class, but it's an abstract method, so the derived classes _must_ implement their own particular version. My reasoning was - we need to enforce a method we know a name of so that `md.sample` can call it. That is `stop_md`, which by default just stops the dynamics on the first failure, but uses the particular `check_if_atoms_ok` (better names welcome). Derived classes can of course re-implement `stop_md` too to have something more particular.

bernstei: OK, I see - the essentially the default behavior is to stop on the first failure, and the derived classes can override that by reimplementing the actual stop function. I'm fine with that, but let me tweak the docstrings so I think that behavior is clear, and I also want to change the function names.

wfl/generate/md/__init__.py
bernstei: More general, but problematic if the trajectory is very long.  I think it'd be easy enough to call one function that returns True/False mid trajectory, and another that returns the entire selected set at the end.  

We could wait until we have a specific use case, but it's so simple, I don't think it's an issue to add both function calls now. Do we want a simple function, or an object like the failure test?

tests/test_md.py
bernstei: Yeah, it's just bad practice. What do you think about passing the MD step #, an integer, instead of using the time, for the example?

bernstei: I added a new info["MD_step"] and used it in a revised version of the test.

wfl/generate/md/__init__.py
bernstei: I generalized it to `traj_select_during_func` and `traj_select_after_func`. See my upcoming commit.

wfl/generate/md/abort.py
bernstei: Fixed in the upcoming commit.

bernstei:
@gelzinyte what do you think of my latest commit and how it addresses the open coversations?

tests/test_md.py
gelzinyte: of course, thanks for the solution.

wfl/generate/md/__init__.py
gelzinyte: I think that's a very good solution, thanks

bernstei:
I'm happy for you to merge whenever you're ready.

PR #142, opened 2022-08-30
Minimahopping

simonwengert:
For the generation of structures this code adds a minima hopping module (including corresponding pytests for the functionality and changes for the appearance in the workflow documentation under https://libatoms.github.io/workflow/wfl.generate.html)


simonwengert:
According to the workflow documentation for creating auto-parallelized functions they should be `a function that takes, normally as its first argument, an iterable, e.g. a list of ase.atoms.Atoms, and returns a list of ase.atoms.Atoms **of the same length**.` 
In the current version this is not necessarily true. It can be a lower number in case the minima-hopping fails for one or more input structures. 
We could force it to have the same length as the input configs, e.g. by using the original input structure for failed minima-hoppings and use the Atoms.info['config_type'] to point this out (and distinguish from successful optimizations). What do you think?

bernstei:
The autopara framework wants to maintain a one-to-one relationship between input and output configs, so yes, it has to return a list of the same length. The right way to return "nothing" for one input config is to return `None` as one of the list items. It may be a good idea to be able to disable that (I can imagine it not making sense for selection operations, for example), but for the moment it's required.

One caution is that missing configs are a poorly tested feature of `ConfigSet`, but in principle should be supported, so let's do this one the way it _should_ work, and (sooner or later) make sure there's a test of that behavior.

simonwengert:
The code has been updated such that also 'None' elements are appended to the list of atoms objects. 

Related to this, I notices that the output returned by the run() function does not contain None elements. They are filtered before entering the outputspec object (e.g. in ../wfl/autoparallelize/pool.py line 140-145). In the (unlikely) case of having only 'None' elements the auto-para framework returns an empty ConfigSet(). To account for this situation I also adjusted the pytests.

bernstei:
Thanks.  There's a large rewrite of ConfigSet/OutputSpec that will probably affect things like handling of missing configs, so we'll have to revisit lots of things like this when that one is finally merged.

bernstei:
Other than that possibility, I'm happy to merge.

simonwengert:
Unfortunately, there seems to be no direct way to control the randomness. In the minimahopping implementation in ASE momenta are initialized randomly without giving the possibility to control the random-generation (and its seed).

bernstei:
Is everyone happy with merging now?

bernstei:
I created #148, and I think I have a fix. I'll push it soon.

simonwengert:
> Is everyone happy with merging now?

Yes, it can be merged now.

PR #146, opened 2022-09-15
Test conda h5py

bernstei:
Test conda h5py, which is failing in another PR.


bernstei:
enough info to open h5y feedstock issue

PR #150, opened 2022-09-16
Autopara info defaults

### LINKED ISSUE #147: Generalize Autoparallelize inputs
As mentioned [here](https://github.com/libAtoms/workflow/pull/121), running certain functions in parallel does not work in an intuitive manner. A common occurence is that the function you're running gives the output:
```
Running wfl.generate.md.sample_autopara_wrappable with num_python_subprocesses=4, num_inputs_per_python_subprocess=None
```
Thus not running anything in parallel, but in series.

I'm guessing the most straightforward solve would be to default the ```num_inputs_per_python_subprocesses``` to the most logical integer, depending on the length of the input iterable and the ```num_python_subprocesses```.

bernstei:
Fix setting default value of `num_inputs_per_python_subprocess`.  Confusion was due to default values in `_autoparallelize_ll`, which have been removed, so also update remaining uses of that function to new `autoparallelize`.

May not pass all tests as of this commit.

closes #147


bernstei:
This is passing all tests here, except for the known broken phono3py. I'm still making sure that my more complete list of tests run locally on my machine all pass.

bernstei:
@gelzinyte @simonwengert @Zausinator does this look OK?

bernstei:
This seems to pass all of my tests. Does someone want to test it, or should I just go ahead and merge this one, and we'll see what happens when the other PRs incorporate the latest changes to main?

wfl/calculators/generic.py
```diff
@@ -13,6 +13,10 @@
 def run_autopara_wrappable(atoms, calculator, properties=None, output_prefix='_auto_', verbose=False, raise_calc_exceptions=False):
     """evaluates configs using an arbitrary calculator and store results in SinglePointCalculator
 
+    Defaults to wfl_num_inputs_per_python_subprocess=10, to avoid recreating the calculator for
```
gelzinyte, line 16: Can this be overwritten in a particular instance? If so, it would be clearer to add `def_`/`default_` to `wfl_generic_num_inputs_per_python_subprocess`?

gelzinyte:
Looking over the changes looks good to me. I'll open issues/PRs if anything comes up after the merge.

wfl/calculators/generic.py
bernstei: It can, I believe, although it might be a good idea to test, given our problems with the default interface.  I guess I do agree that it's worth the even more verbose name.  I'll update that, and try to make a test of that functionality.  I think I'll have to do something like capture stdout and make sure the correct message is printed.

bernstei: please resolve this conversation if you think my latest commit addresses the issue. If not, let's discuss further.

PR #156, opened 2022-09-21
ace_fit routines looks for ace_fit.jl by asking julia ACE1pack for its install path

### LINKED ISSUE #155: automatically find ace_fit.jl
I have a little shell script that can find the location of `ace_fit.jl`, assuming that `julia` is executable.  Is it worth adding that to the ace fitting routines?

bernstei:
If `WFL_ACE_FIT_COMMAND` not specified, run `julia` (or `WFL_JULIA_COMMAND`) to find out path to `ACE1pack` which will be used.

As a side effect, requires ace_fit tests to have an explicit `--runacefit` marker instead of the presence an env var with the fitting script path.  Refactored the whole pytest marker/runtime flag routines, since they were getting out of control.

closes #155


gelzinyte:
Works like a charm! (on Womble and locally, on development as well as freshly installed branches)

wfl/fit/ace.py
```diff
@@ -258,11 +258,12 @@ def run_ace_fit(fitting_configs, ace_fit_params, skip_if_present=False, run_dir=
 
     if ace_fit_command is None:
         if "WFL_ACE_FIT_COMMAND" in os.environ:
-           ace_fit_command = os.environ["WFL_ACE_FIT_COMMAND"]
+            ace_fit_command = os.environ["WFL_ACE_FIT_COMMAND"]
         else:
-            raise ValueError("Executable and path to ace_fit.jl must be given either as a `ace_fit_command` argument "\
-                "to `run_ace_fit()` or via WFL_ACE_FIT_COMMAND environment variable ")
-    
+            julia = os.environ.get("WFL_JULIA_COMMAND", "julia")
```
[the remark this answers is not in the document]
bernstei: Where `wfl.utils.julia` ? And would you just want a function that returns the `julia` executable, or something that handles running the julia command line?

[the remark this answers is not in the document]
bernstei: I just did the simplest thing in my latest commit. Do you think that's enough?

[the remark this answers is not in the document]
gelzinyte: sorry I wasn't clear enough! I meant the path to the `ace_fit.jl` (with or without `julia` bit). I'll add that on.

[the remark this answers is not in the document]
gelzinyte: I've added the change I had in mind, let me know how that looks.

wfl/utils/julia.py
```diff
@@ -1,4 +1,12 @@
 import os
+import subprocess
+import shlex
+from pathlib import Path
 
 def julia_exec_path():
     return os.environ.get("WFL_JULIA_COMMAND", "julia")
+
```
bernstei, line 8: I like this in a function, but do you really think it belongs outside of `fit/ace.py`?  It's entirely specific to the fitting code. The julia interpreter I can see being useful for other things, but not the fitting script itself.

bernstei: Sorry - I find the github interface for associating comments with specific source lines confusing (or just broken).

I mean the function that returns the `run_train.py` path - don't you think that should live in `fit/ace.py`, or maybe `fit/ace_utils.py`, or even `fit/utils.py`?

gelzinyte: yeah, that sounds more reasonable. I moved it to `fit/utils.py`.

bernstei: I'm going to modify it further, making `ace_fit_jl_path` look up its own julia path, and add it to the command, so the calling function can do it in one step.  Just running the tests manually now, to make sure it's still OK.

gelzinyte: yes, that's how it should be 😄

bernstei:
OK - I'll merge as soon as my local tests finish successfully. Maybe I'll add julia to the CI.

bernstei:
OK - a ridiculous amount of fighting with julia later (much of which caused by things I really should realized more quickly), it's working.

PR #163, opened 2022-10-05
weight_from_sigma in ACE fit params rather than argument to ACE fitting wrapper

bernstei:
Remove weight_from_sigma as an argument to ace fit wrapper, instead make it a special value in weights dict in params

### DIFF
--- a/wfl/fit/ace.py
+++ b/wfl/fit/ace.py
@@ -21,7 +21,7 @@ from wfl.fit.utils import ace_fit_jl_path
 from expyre import ExPyRe
 import wfl.scripts
 
-def fit(fitting_configs, ACE_name, ace_fit_params, ace_fit_command=None, weight_from_sigma=False,
+def fit(fitting_configs, ACE_name, ace_fit_params, ace_fit_command=None,
         ref_property_prefix='REF_', skip_if_present=False, run_dir='.', dry_run=False,
         verbose=True, remote_info=None, remote_label=None, wait_for_results=True):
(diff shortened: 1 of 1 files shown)


wfl/fit/ace.py
```diff
@@ -132,14 +128,17 @@ def prepare_params(ACE_name, fitting_configs, ace_fit_params, run_dir='.', ref_p
...
 
-    if weight_from_sigma:
-        ace_fit_params["weights"] = {}
-        for at_i, at in enumerate(fitting_configs):
-            config_type = f"config_{at_i}"
-            at.info["config_type"] = config_type
-            ace_fit_params["weights"][config_type] = { "E": 1.0 / at.info.get("energy_sigma", 1.0),
-                                                       "F": 1.0 / at.info.get("force_sigma", 1.0),
-                                                       "V": 1.0 / at.info.get("virial_sigma", 1.0) }
+    from_sigma = ace_fit_params.get("weights", {}).get("from_sigma")
```
gelzinyte, line 131: Maybe would be good to mention this behaviour in docstring of this function or someplace else?

bernstei: Yes, good point. I'll update the docstring.

bernstei: addressed in bba5be1

bernstei:
@gelzinyte let me know if you think it's OK now, otherwise feel free to merge once the tests pass (or I can do it).

PR #165, opened 2022-10-12
Ace fit params with yaml

bernstei:
save ace fitting parameters for `ace_fit.jl` as a yaml, not a json, so integer dict keys will not be converted to strings.


wfl/fit/ace.py
```diff
@@ -132,14 +129,17 @@ def prepare_params(ACE_name, fitting_configs, ace_fit_params, run_dir='.', ref_p
...
 
-    if weight_from_sigma:
-        ace_fit_params["weights"] = {}
-        for at_i, at in enumerate(fitting_configs):
-            config_type = f"config_{at_i}"
-            at.info["config_type"] = config_type
-            ace_fit_params["weights"][config_type] = { "E": 1.0 / at.info.get("energy_sigma", 1.0),
-                                                       "F": 1.0 / at.info.get("force_sigma", 1.0),
-                                                       "V": 1.0 / at.info.get("virial_sigma", 1.0) }
+    from_sigma = ace_fit_params.get("weights", {}).get("from_sigma")
```
gelzinyte, line 132: Is this documented somewhere? I think it would be good to briefly mention?

gelzinyte: Maybe just in a docstring of this function?

gelzinyte:
Looks good I think. The only other thing to watch out for is the params entries that in Julia are of tuple-to-tuple type. But those are a bit rarer and we don't modify them in the workflow fitting script, so I think we can leave it for users to take care of. 

Just to mention, if these tuples are saved (from python or other to yaml) as strings (e.g. `{"(C, C)" : "(1.2, 4)"}`), these get automatically parsed appropriately ( to `Dict((C, C) => (1.2, 4))`) in `ACE1pack.fill_defaults()`, at least when read from yaml.

wfl/fit/ace.py
gelzinyte: Never mind - left a comment in the original PR.

bernstei:
@gelzinyte what do you think about merging this one, since it's clearly safer than using JSON, regardless of whether ACE actually gets automatic conversion of integer-as-string keys?

bernstei:
I guess since JSON didn't handle tuple keys it was already required that they be encoded as strings, right? We could fix that, too, and in my numpy-removing routine turn tuples into strings?

bernstei:
In the julia dict format, e.g. `Dict((C, C) => (1.2, 4))`, is it really supposed to be that, or actually `Dict(("C", "C") => (1.2, 4))` ?


If the latter (and julia accepts strings in single quotes) I can just do `if isinstance(item, tuple): return str(tuple)` (which would return a string containing `('C', 'C')`), but if the former, it'd have to be something like `return "(" + ", ".join([str(subitem) for subitem in item]) + ")"` so it returns `(C, C)`

gelzinyte:
yeah, can merge this. 

You're right, it's actually `Dict(("C", "C") => (1.2, 4))` in Julia. I think the problem was with serialising nested quotes, but julia also doesn't accept singly-quoted strings. I think let's not worry about this now...

bernstei:
It was easy enough to implement.  I guess I'll add a test.

bernstei:
OK, I got the tests fixed, and the right format for exporting tuples so that julia will read them correctly. I'm ready to merge, unless you have any other concerns.

Issue #149, opened 2022-09-16
traj_select_during_func in wfl/generate/md/__init__.py not working as expected


simonwengert:
When running an md via the `sample()` function in combination with using the `traj_select_during_func` option, the last image of an MD does not have to pass the `traj_select_during_func` function. Thus, for instance with a selection-function that will not less any atom pass, the `sample()` function will still return atoms-objects. Tested for instance with the following function
```
def traj_select_during_func(atoms):
    return False
```
I believe the issue is related to this code block in `wfl/generate/md/__init__.py` (Line 224-226)
```
if len(traj) == 0 or traj[-1] != at:
    at.info['MD_time_fs'] = cur_step * dt
    traj.append(at_copy_save_results(at, results_prefix=results_prefix))
```
where the `traj[-1] != at` is very likely to be fulfilled when using  `traj_select_during_func`.

A possible solution would be 
```
if len(traj) == 0 or traj[-1] != at:
    if traj_select_during_func is None or traj_select_during_func(at):
        at.info['MD_time_fs'] = cur_step * dt
        traj.append(at_copy_save_results(at, results_prefix=results_prefix))
```

bernstei:
Do we agree that MD sampling should always return at least the last configuration? Or if a user passes a trajectory selection function that thinks nothing should be returned, it should return nothing?

simonwengert:
I think it would be more intuitive to let the function return nothing if the specified criterion of the selection function is too tight. (And the user might better rethink the specified criterion in such cases)

bernstei:
I tend to agree.  Can you create a PR with your proposed solution?

BTW, what do you think of having the function parameter being `traj_select_during_func=lambda at: True` instead of `None`, so we don't have to test for it being `not None`?

simonwengert:
> BTW, what do you think of having the function parameter being `traj_select_during_func=lambda at: True` instead of `None`, so we don't have to test for it being `not None`?

I like the idea.

PR #167, opened 2022-10-18
in generate.md.sample() via the traj_select_during_func argument, require last image pass this filter too.

simonwengert:
Fixes issue #149

### DIFF
--- a/tests/test_md.py
+++ b/tests/test_md.py
@@ -142,14 +142,15 @@ def test_subselector_function_during(cu_slab):
 
     calc = EMT()
 
-    inputs = ConfigSet(cu_slab)
-    outputs = OutputSpec()
+    for steps in [300, 301]:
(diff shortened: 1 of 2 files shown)


bernstei:
Changes look good, but could you add the change to the default to lambda, as mentioned in https://github.com/libAtoms/workflow/issues/149#issuecomment-1282387140 ?

Also, could you add a test that confirms that the last image is correctly included/excluded, probably by modifying one of the existing tests of the sampling selection behavior?

simonwengert:
Do you think the doc-string for the lambda-default of the traj_select_during_func parameter is clear?

bernstei:
> Do you think the doc-string for the lambda-default of the traj_select_during_func parameter is clear?

Looks good to me.  @gelzinyte do you have any suggestions, or should I go ahead and merge?

Issue #168, opened 2022-10-18
Online documentation does not display `docs/source/wfl.generate.md.rst`


simonwengert:
When generating the html *circular toctree references* are detected. I suspect the *.md extension causes some confusion here.

bernstei:
Yes - I opened this https://github.com/sphinx-doc/sphinx/issues/10914

I don't see any way to fix this without renaming our md module or changing sphinx internals.

gelzinyte:
I think this is fixed now?

bernstei:
Yes, closing now.  If #169's description had the line `closes #168` it would have happened automatically.

PR #169, opened 2022-10-18
fix circular toctree references (probably caused by confusion from *.md extension)

simonwengert:
Fixes  Issue #168


bernstei:
I have an idea  - those file are semi-automatically generated by `sphinx-apidoc -o source ../wfl`. It _is_ possible to deal with this semi-automatically (unlike what I wrote in the issue comments) by making this command in the docs `Makefile`, and adding `mv`/`sed` commands to rename the file and its reference. 

@simonwengert would you be able to do that, or should I close this PR and open a different one for my semi-automatic solution?

simonwengert:
You meant something like that (see PR)?

bernstei:
> You meant something like that (see PR)?

That's the right `sed`, command but right now it's modifying a github committed file.  I meant that all the files generated by the `sphix-apidoc` command need to be removed from the repo and instead generated dynamically by the `docs/Makefile` rule before the `sed`.

bernstei:
replaced by #176

PR #177, opened 2022-10-25
Better handling of remote job failure

### LINKED ISSUE #61: better handling of (multiple) failed DFT calculations in a big list of jobs
Right now if a DFT job fails, recovering is ugly.  Running the operation again will note the failure during the result gather stage, but it will (by default at least) exit immediately.  It will therefore not notice any other failures, and even if that failure is cleaned up manually, rerunning the workflow will resubmit that job, then get to the gather stage, and will be waiting for that job to finish before it even notices the later failures.

It would be better if the pipeline multiple remote jobs just kept track of all the failures, and reported all of them at the end of the loop.  The only downside is that it will not be able to raise the same exception (necessarily), because different failures might have occurred due to different exceptions.

bernstei:
Replace `RemoteInfo.skip_failures` with `RemoteInfo.ignore_failed_jobs`, and add `RemoteInfo.resubmit_killed_jobs`.  Fix remote job tests, and add tests for various job-failure handling behaviors.

closes #61


bernstei:
The failed tests appear to be an intermittent randomly returned error status 143 from the python process doing the remote run for `tests/test_remote_run.py::test_resubmit_killed_jobs`.  I'm looking into what's causing them.

bernstei:
> The failed tests appear to be an intermittent randomly returned error status 143 from the python process doing the remote run for `tests/test_remote_run.py::test_resubmit_killed_jobs`. I'm looking into what's causing them.

These are currently too intermittent to debug, but perhaps the additional stdout/stderr messages from ExPyRe (https://github.com/libAtoms/ExPyRe/pull/26 and https://github.com/libAtoms/ExPyRe/pull/27) will help next time they show up.

It does, however, bring up a separate question. Right now the only circumstance under which a remote job will be restarted (if you set `restart_killed_jobs`) is if it gets a `ExPyReJobDiedError`, which only happens when it stops running before it's written either a success or error status file. This normally happens when the queuing system kills it because it ran out of time, but maybe also if it runs out of memory. Do we instead want to (always, or controlled by a flag) restart any remote job that got an exception? I can imagine, for example, that the out-of-memory (OOM) killer would send a signal to the python process, which would result in a `RuntimeError`, I think.  That wouldn't get restarted automatically, even though you could try to fix it by increasing the number of nodes or requesting more nodes per job.  Or do we want for evidence that this happens in practice before we make the interface more complex?

gelzinyte:
I would be up for adding this once we run into such problems

gelzinyte:
The code looks good to me. I'll try running jobs from this branch over next week. So can keep this open meanwhile or merge now and fix with new PRs if any bugs come up.

bernstei:
> The code looks good to me. I'll try running jobs from this branch over next week. So can keep this open meanwhile or merge now and fix with new PRs if any bugs come up.

I'll wait for another week, since I don't think anyone urgently needs it in main, and it'll be cleaner if there are any bugs to fix them in this PR.

bernstei:
See also #180

bernstei:
@gelzinyte any progress testing this branch?

bernstei:
I've been using it happily, so I'm going to merge, and if someone has an issue, we can reopen or open a new one.

Issue #53, opened 2022-03-22
buildcell


bernstei:
Replace prep stage writing files to return strings, workflow can write them if that makes sense.

Maybe remove separate prep stage, if it can be done cheaply enough, and do it "on the fly" during actual running of buildcell.

PR #178, opened 2022-10-27
Added example on generating random structures via buildcell

simonwengert:


gelzinyte:
Thanks for the example! 

Would you mind adding link(s) pointing to `examples.buildcell.md` in `examples.index.md` under appropriate categories? I think an index would help people find most relevant examples, even if this one is quite self explanatory.

simonwengert:
The current example is based on the version of `create_input()` that writes to file. I agree with what you mentioned in Issue #53 and might adjust the example if PR #179 is fine for you.

bernstei:
Did we ever manage to come up with a system so that all our examples are actually runnable, so we can write tests to ensure that they remain correct?

simonwengert:
I don't think we have one. Would be good to have such a system though. Are you aware of tools that can be useful for this? My first idea would be to write a small parser-function that extracts the code blocks from the files as strings which can then be executed in a second step.

simonwengert:
I am thinking about an additional file in our test suit ```test_doc_examples.py``` which could look like this:

`import os
import pytest

def _get_coding_blocks(filename):
    "Parse ```filename``` for coding blocks and return as list of strings."
    with open(filename, 'r') as fo:
        lines = fo.readlines()

    coding_blocks = []
    in_codeblock = False

    for l_i in lines:

        if l_i.startswith('```'):
            if not in_codeblock:
                coding_block = ''
            else:
                coding_blocks.append(coding_block)
            in_codeblock = not in_codeblock
            continue

        if in_codeblock:
            coding_block += l_i

    return coding_blocks


@pytest.fixture
def basepath():
    return os.path.join(os.path.dirname(__file__), '../', 'docs', 'source')


def test_examples_dimers(basepath):
    coding_blocks = _get_coding_blocks(f'{basepath}/examples.dimers.md')
    exec(coding_blocks[0])
    assert os.path.exists('dimers.xyz')

 
def test_examples_buildcell(basepath):
    coding_blocks = _get_coding_blocks(f'{basepath}/examples.buildcell.md')
    exec(coding_blocks[0])
    assert os.path.exists('buildcell_output.xyz')
`

bernstei:
What about making the examples jupyter notebooks?  Those can be turned into pure code (to run for tests) and also into html with `jupyter nbconvert` ?

simonwengert:
> What about making the examples jupyter notebooks? Those can be turned into pure code (to run for tests) and also into html with jupyter nbconvert.

There is a sphinx-extension _nbspinx_ that seems to be quite convenient for obtaining the online documentation from ipynb-files. I tested it and it seems to work.

bernstei:
@gelzinyte @simonwengert should we try out this notebook-based approach on this buildcell example?  Or is that still something we want to defer to an issue, to be dealt with later?

simonwengert:
The buildcell example has now been turned into a jupyter notebook. To make the html generation work **nbsphinx** and **pandoc** needed to be installed.

I wrote the test for it in a general way such that additional tests for other examples can be readily appended to @pytest.mark.parametrize()

bernstei:
> To make the html generation work nbsphinx and pandoc needed to be installed.

I assume we don't want those as requirements for the package, but only installed by the github autodoc CI and users who want to be able to build their own docs?

simonwengert:
I think so too. This probably means we should set a flag or something to the tests such that they are only executed on demand.

BTW I needed to install pandoc via `conda install -c conda-forge pandoc`. (I guess to version `pip` provides is too old.)

bernstei:
Why is there anything special with the tests?  I glanced at your code and it seems to be pure JSON parsing of the notebooks. It's only the _docs_ that depend on those packages, right? 

As for conda vs. pip pandoc, that's ugly.  I'd like to avoid conda, as it completely breaks our system. Have you checked the OS-provided version in the github CI?
```
Pandoc is in the [Debian](https://packages.debian.org/pandoc), [Ubuntu](https://packages.ubuntu.com/pandoc), [Slackware](https://www.slackbuilds.org/result/?search=pandoc&sv=), [Arch](https://www.archlinux.org/packages/community/x86_64/pandoc/), [Fedora](https://apps.fedoraproject.org/packages/pandoc), [NiXOS](https://nixos.org/nixos/packages.html), [openSUSE](https://software.opensuse.org/package/pandoc), [gentoo](https://packages.gentoo.org/package/app-text/pandoc) and [Void](https://voidlinux.org/) repositories.
```

bernstei:
Thoughts about how to make the buildcell example and failing test be robust to the choice of installation location for the `buildcell` executable?  Should we add its location to `PATH` in the CI workflow, so the notebook can just refer to it as `buildcell`? Right now the test for the python API buildcell wrapper uses a specific `WFL_PYTEST_BUILDCELL` env var, which is set by the github CI.

docs/source/examples.buildcell.ipynb
```diff
@@ -0,0 +1,71 @@
...
+    "from wfl.generate.buildcell import create_input, run\n",
+    "from wfl.configset import OutputSpec\n",
+    "\n",
+    "\n",
+    "inputs = range(20)\n",
+    "outputspec = OutputSpec('buildcell_output.xyz')\n",
+    "\n",
+    "buildcell_input = create_input(z=13, vol_per_atom=10, bond_lengths=2)\n",
+    "\n",
+    "run(inputs=inputs,\n",
```
bernstei, line 40: Since this is an example, as well as a test, I'd like to strongly encourage "pipeline" syntax, i.e. `outputs = run(inputs, ...)`, so it's more `OutputSpec`-detail independent, and it  automatically gives a `ConfigSet` with the created configs.  It also makes it easier to add functionality tests list `assert len(list(outputs)) == 10`.

docs/source/examples.buildcell.ipynb
```diff
@@ -0,0 +1,71 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4fd84da0-9483-4a01-9aa6-8de6d53aed9d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "from wfl.generate.buildcell import create_input, run\n",
```
bernstei, line 31: Do you think it'll be more readable if it's
```
from wfl.generate import buildcell
.
.
outputs = buildcell.run(inputs, outputspec, ...)
```
I'm reconsidering my choice to name the actual operations very nondescript things like `run`.

bernstei:
Generally looks very nice, just a couple of comments on the syntax.

simonwengert:
> As for conda vs. pip pandoc, that's ugly. I'd like to avoid conda, as it completely breaks our system. Have you checked the OS-provided version in the github CI?

I tested the build-docs workflow using the apt-get installation of pandoc and it seems to work fine.

> Thoughts about how to make the buildcell example and failing test be robust to the choice of installation location for the buildcell executable? Should we add its location to PATH in the CI workflow, so the notebook can just refer to it as buildcell? Right now the test for the python API buildcell wrapper uses a specific WFL_PYTEST_BUILDCELL env var, which is set by the github CI.

That's that a good idea. I adjusted the pytests.yml accordingly. This should fix the failing CI test suit. If someone runs the tests locally and `buildcell` is not in `PATH`, however, the test will still fail since we don't test for WFL_PYTEST_BUILDCELL as in test_buildcell.py. So I added a pytest.mark to skip in this case.

> Generally looks very nice, just a couple of comments on the syntax.
Thanks for the comments. I agree with your suggestions and refined accordingly.

tests/test_doc_examples.py
```diff
@@ -12,7 +13,8 @@ def _get_coding_blocks(nb_file):
 @pytest.mark.parametrize(
     ('nb_file', 'idx_execute'),
     (
-        pytest.param('examples.buildcell.ipynb', 'all', id='buildcell'),
+        pytest.param('examples.buildcell.ipynb', 'all', id='buildcell',
```
bernstei, line 16: The existing `buildcell` test uses a different env var, but I agree that there's no easy way to make this example test use it. Can you change the existing `buildcell` test to fall back to searching in PATH (like this example test) if `WFL_PYTEST_BUILDCELL` isn't set?

bernstei:
Looks good to me. Just a little change to make the explicit buildcell pytest consistent in terms of finding the executable.

tests/test_doc_examples.py
simonwengert: I changed the tests to look for both WFL_PYTEST_BUILDCELL and in PATH.

When I was looking into the buildcell-tests, I was wondering about the remote case since the path to the buildcell-executable on the cluster might differ from the one on a local machine. Is this something to worry about?

When running it via `pytest test_buildcell.py --runremote --basetemp $HOME/pytest` the tests run and pass on my local machine, however, jobs are also not send to the cluster. Is this an intended behavior?

bernstei: > When I was looking into the buildcell-tests, I was wondering about the remote case since the path to the buildcell-executable on the cluster might differ from the one on a local machine. Is this something to worry about?

In some sense yes, but given that the remote machine can be arbitrarily different from the local machine, there's not going to be a general solution.  I'm happy with requiring that either the absolute path be the same, or that it be in `PATH` (and therefore only need the executable name). If needed we can add support for a pytest-wide customization of the remote job `PATH` (e.g. by having pytest add an `env_var` for all its remote jobs).

I think it'd be too hard to make the system be entirely agnostic, in a completely general way, as to what machine it's running on. I can live with the current situation, i.e. that if the user specifies a remote job, they have to make sure the paths they pass are the correct ones for the remote machine.  It's not optimal for testing, but I think that for real production runs, if you're using a script you can always make the path to things like buildcell come from  a variable, which you can set at runtime.

bernstei:
I'm ready to merge, unless someone has thoughts about the open conversation about local vs. remote buildcell paths.

tests/test_doc_examples.py
simonwengert: I see. Thanks for clarifying.

simonwengert:
I think we can merge it now.

PR #179, opened 2022-10-27
for generation via buildcell, let prep stage return a string and only optionally write to file

simonwengert:
see Issue #53

### DIFF
--- a/wfl/generate/buildcell.py
+++ b/wfl/generate/buildcell.py
@@ -16,6 +16,47 @@ from wfl.utils.round_sig_figs import round_sig_figs
 # and bond len as values
 def create_input(z, vol_per_atom, bond_lengths, composition=None, RSS_min_vol_factor=0.5,
                  vol_range=(0.95, 1.05), min_sep_factor=0.9, symmops='1-8', natom=(6, 24), odd=None, verbose=False):
+    """
+    Create an input string suitable for a random structure search (RSS) via AIRSS' `buildcell`.
+
+    Parameters
+    ----------
+    z: int / list(int)
+        Atomic number(s) to be present in the system.
+    vol_per_atom: float / list(float)
+        Volume to be reserved per atom in the system
+        (thus, affects overall unit cell volume/system density).
+    bond_lengths: float / list(float)
+        Defines the separation between atoms.
+    composition: list(int), default None
+        Defines the stoichiometric ratio between species,
+        in case multiple species are present (i.e. `z` is a list)
+    RSS_min_vol_factor: float, default 0.5
+        Together with `vol_per_atom` defines the minimum volume
+        to be reserved per atom.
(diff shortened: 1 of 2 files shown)


bernstei:
I'd actually suggest getting rid of the filename, and let whatever calls it write the string to a file if they want. That's been on my list of things to do for the refactors gap-rss anyway.

bernstei:
Change looks good, but now I see that there's no docstring.  As long as this is in your repo I can't write one, so you'll have to, or we could create a branch in the local repo that's cloned from yours.

simonwengert:
Right, also noted that. I added a doc-string now.

wfl/generate/buildcell.py
```diff
@@ -16,6 +16,47 @@
 # and bond len as values
 def create_input(z, vol_per_atom, bond_lengths, composition=None, RSS_min_vol_factor=0.5,
                  vol_range=(0.95, 1.05), min_sep_factor=0.9, symmops='1-8', natom=(6, 24), odd=None, verbose=False):
+    """
+    Create an input string suitable for a random structure search (RSS) via AIRSS' `buildcell`.
```
bernstei, line 20: sphinx-apidoc works better with two backquotes, so \`\`buildcell\`\`

wfl/generate/buildcell.py
```diff
@@ -16,6 +16,47 @@
...
+        between atoms.
+    symmops: str, default '1-8'
+        Build structures have a specified number of symmetry operations.
+        For crystals, the allowed values are (1,2,3,4,6,8,12,16,24,48).
+        For clusters (indicated with #CLUSTER), the allowed values are
+        (1,2,3,5,4,6,7,8,9,10,11,12,24). Ranges are allowed (see default).
+    natom: tupe(2), default (6, 24)
+        Defines the range of total numbers of atoms to be present
+        in build structures.
+    odd: str, default None
```
bernstei, line 50: There's also an `"also"` value.  I suggest
```
odd: str, default None
    control odd/even numbers of atoms in cell. Use ``"only"`` to only produce cells with 
    odd numbers of atoms, ``"also"`` to produce both odd and even, and ``None`` for even only.
```

bernstei:
Thanks @simonwengert .  Looks good, except for some tiny wording changes (see in-code comments above).  It'd be nice to have a test, I guess added to `tests/test_buildcell.py`.  If you feel like doing that, great, if not, we should open an issue.

bernstei:
Looks good now.  Just let me know if I should merge now, or if you're working on a test.

simonwengert:
@bernstei I'm writing on a test. Actually, I've just finished writing it and will push it now.

bernstei:
Fancy.  Yes, that looks good.  @simonwengert if you think it's ready, I'll merge.  If you think it needs another look, let me know.

simonwengert:
I would think it's ready to be merged

Issue #58, opened 2022-03-22
refactor cli.py into separate source files


bernstei:
`cli.py` is too long, and also includes some things that probably don't belong in it at all.  Refactor it to make the source files more modular, and deprecate or get rid of unneeded bits.

Zausinator:
Is the goal to have ```cli.py``` be the center for all wrapper functions within wfl? This would potentially make working with a python API a lot more streamlined. For instance, a wrapper function that automatically generates ```ConfigSet``` and ```OutputSpec``` instances for given in- and outputs would shorten our scripts a lot.

bernstei:
I don't think that's what we envisioned, but it's worth thinking about. Do you have a specific idea of how this might work?

There are two wrappings to be done - the autoparallelize (which is somewhat cleaner if you use `iloop`, although the lack of native python syntax keeps it from being entirely clean), and the command line parsing.  The latter, especially, I don't think can be automated because there's such a variety of things that the input strings need to be transformed into, although certainly having something standardized for input and output file arguments (whether just a convention or an actual decorator) and then turns those into `ConfigSet`/`OutputSpec` objects would be very nice.

gelzinyte:
Functions in the current `cli.py`: 
- keep SMILES -> xyz (useful for Elena) 
- remove gather & strip: without ABCD it does what ase convert does. Potentially re-introduce if ABCD is resurrected 
- remove `committee`. Maybe think about how to introduce that via generic - e.g. list of calc initializers, etc. See issue ...
- keep `repeat-buildcell` - seems generic enough, already implemented and potentially useful, so let's just not remove it. 
- keep cur. Also add local CUR, see issue ...
- keep calculating descriptor
- MLIP calculators: separate for `wfl calculate gap`, ace, mace. Anything else? 
- DFT calculators: no cli support, python interface only. 
- reference-error: refactor, only do errors (from file that has reference and predicted E/F already). Issue #171 
- `simple-gap` remove (doubles up to `gap_fit`, not really useful); similarly to `multistage-gap`.

bernstei:
Bottom line: keep only
- SMILES -> xyz
- buildcell
- CUR select (add local)
- descriptor calc
- MLIP calc, separate command for each
- reference error (using refactored #171)

create cli utils:
- standardize `ConfigSet` and `OutputSpec` input arguments

gabor1:
allow all kinds of selections, not just CUR ? (we had this general selection idea via a lambda function)

[...]

PR #181, opened 2022-11-01
Refactor cli

gelzinyte:
Closes issue #58 

Will implement `wfl error` once PR #174 is merged. 

Most of the code invoked by the cli commands is copied from the previous interface - what should be updated here?


wfl/cli/cli_options.py
```diff
@@ -1,12 +1,43 @@
...
+        return key_val_str_to_dict(value)
+    else:
+        return {}
+
+def extra_info(f):
+    """Parse key=val string and return a dictionary"""
+    f = click.option("--extra-info", "-i", callback=_parse_extra_info, help="Extra key=val pairs to add to Atoms.info")(f)
+    return f
+
+def descriptor(f):
```
bernstei, line 37: Is there a use case for putting this in a centralized location, as opposed to right where the descriptor calc CLI is defined?  I don't obviously see what other command might need a standardized `--descriptor` flag.

wfl/cli/commands/eval.py
```diff
@@ -0,0 +1,81 @@
+import click
+
+from quippy.potential import Potential
+
+from wfl.autoparallelize.autoparainfo import AutoparaInfo
+from wfl.cli import cli_options as opt
+
+
+@click.command("gap")
+@click.option("--gap-fname", '-g', type=click.Path(), help="Path to GAP XML")
```
bernstei, line 10: I think I'd prefer for the parameter file name argument to be the same for all the potentials.  In fact, all of these options (except the extra MACE ones) can and should be standardized by centralizing them for all the ML evaluation subcommands.

wfl/cli/commands/select.py
```diff
@@ -0,0 +1,33 @@
...
+    """Select structures by CUR"""    
+
+    wfl.select.by_descriptor.CUR_conf_global(
+        inputs=inputs,
+        outputs=outputs,
+        num=n_configs,
+        at_descs_info_key=key, kernel_exp=kernel_exponent, stochastic=not deterministic,
+        keep_descriptor_info=keep_descriptor)
+
+
```
bernstei, line 30: Would a simple interface to the per-config bool function be useful here?  We could do it a bit like `ase convert`, where the user specifies a string that evaluates to a bool for a standardized name, `atoms`, and we can use `eval("lambda atoms: " + user_string)` and pass that to the selection function.

wfl/cli/commands/eval.py
gelzinyte: done

wfl/cli/commands/select.py
gelzinyte: I think that'll be useful, let me add it in.

gelzinyte:
I left out unit testing for that and other functions that I thought are a bit awkward to test command line interface for, but we have unit tests for the python functions. Any you think I should add unit tests for? I'd like to fill in the cli functions for error, etc in a separate PR.

docs/source/examples.normal_modes.md
```diff
@@ -14,14 +14,14 @@ from wfl.configset import ConfigSet, OutputSpec
 from wfl.generate import normal_modes as nm 
 
 mols = [molecule("CH4"), molecule("H2O")]
-ConfigSet = ConfigSet(mols)
+configset = ConfigSet(mols)
 OutputSpec = OutputSpec("molecules.normal_modes.xyz")
```
bernstei, line 18: like `configset`, this one should also not redefine the symbol for the class itself

gelzinyte: slipped through, just fixed.

docs/source/examples.orca_cli.md
```diff
@@ -1,136 +0,0 @@
...
-* `orcablocks` - [ORCA block input](https://sites.google.com/site/orcainputlibrary/general-input?authuser=0#h.k0jcjw8bcgcx), the rest of ORCA input file, excluding the coordinate definition.
-* `output_prefix` - prefix to the results' keys in `atoms.info` and `atoms.arrays`. Here DFT results will be saved to `dft_energy`, `dft_forces` and `dft_dipole`. 
-* `calculator-exec` - path to ORCA executable.
-* `scratchdir` - directory on a SSD for more efficient calculations. 
-* `input` - `.xyz` file to be evaluated. 
-* `output` - `.xyz` to save configs with results to. This command will refuse to overwrite if the file exists. 
-* `--keep-files default` - only `.inp`, `.out`, `.ase`, `.engrad`, `.xyz` and `_trj.xyz` files will be saved, where applicable, upon success. All files will be kept if the calculation failed.
-* `wfl -v ref-method orca-eval` - Workflow command to evaluate DFT, in verbose mode. 
-
-With this command each structure will be evaluated sequentially on a single core.
```
bernstei, line 36: Are you setting `num_python_subprocesses=1` explicitly, or are we relying on `WFL_NUM_PYTHON_SUBPROCESSES` not to be set?

docs/source/index.rst
```diff
@@ -9,7 +9,7 @@ Welcome to Workflow's documentation!
...
 
-The main functions of Workflow is to efficiently parallelise operations over a set of atomic configurations (Atomic Simulation Environment's "Atoms" objects). Given an operation that is defined to act a single configuration (e.g. evaluate energy of a structure with CASTEP ASE calculator), Workflow may apply the operation to multiple configurations in parallel. Workflow also interfaces with [ExPyRe](https://github.com/libAtoms/ExPyRe/tree/main/expyre) to manage evaluation of (autoparallelized) python functions via a queueing system on a (remote) cluster. 
+The main functions of Workflow is to efficiently parallelise operations over a set of atomic configurations (Atomic Simulation Environment's "Atoms" objects). Given an operation that is defined to act a single configuration (e.g. evaluate energy of a structure with CASTEP ASE calculator), Workflow may apply the operation to multiple configurations in parallel. Workflow also interfaces with `ExPyRe <https://github.com/libAtoms/ExPyRe/tree/main/expyre>`_ to manage evaluation of (autoparallelized) python functions via a queueing system on a (remote) cluster.
```
bernstei, line 12: `main functions of` -> `main function of`

Given the generic nature of the word "workflow", what do you think about using "wfl" as the default name?

tests/cli/test_descriptor.py
```diff
@@ -0,0 +1,64 @@
...
+from click.testing import CliRunner
+
+import pytest
+from pytest import approx
+
+from ase import Atoms
+from ase.io import write, read
+
+from wfl.cli.cli import cli
+from wfl.configset import ConfigSet, OutputSpec
```
bernstei, line 14: Do we actually need ConfigSet/OutputSpec for this test?

wfl/cli/cli.py
```diff
@@ -1,56 +1,15 @@
...
-from wfl.configset import ConfigSet, OutputSpec
-from wfl.generate import normal_modes as nm 
-import wfl.generate.smiles
-import wfl.utils.misc
-import wfl.generate.buildcell
-import wfl.select.by_descriptor
-import wfl.descriptors.quippy
-
-from wfl.utils import gap_xml_tools
+from wfl.cli import cli_options as opt
```
bernstei, line 4: Is this actually used in the top level `cli.py`, or only the actual subgroups?

wfl/cli/cli_options.py
```diff
@@ -1,12 +1,52 @@
...
 
+def _to_ConfigSet(ctx, param, value):
+    return ConfigSet(value)
 
-def file_input_options(f):
-    """"""
-    f = argument("inputs", nargs=-1)(f)
-    f = option("--index", "-i", type=STRING, required=False, help="Pass this index to configset globally")(f)
+def inputs(f):
+    """Create ConfigSet for given filename(s)"""
```
bernstei, line 10: This is the docstring for this decorator, right? How about 
```
"""add a standard option for the ConfigSet inputs"""
````
and same for `outputs()`.

File: wfl/cli/commands/descriptor.py [hunk 10]
```diff
@@ -0,0 +1,24 @@
+import click
+from wfl.cli import cli_options as opt
+import wfl.descriptors.quippy
+
+@click.command("quippy")
+@click.pass_context
+@click.option("--local", is_flag=True, help="calculate a local (per-atom) descriptor")
+@click.option("--force", is_flag=True, help="overwrite existing info or arrays item if present")
+@click.option("--descriptor", type=click.STRING, required=True, help="quippy.Descriptor arg string")
```
bernstei, line 9: `quippy.descriptors.Descriptor`, technically

wfl/cli/commands/generate.py
```diff
@@ -0,0 +1,69 @@
...
+    ##EXTRA_INFO <info_field>=<value> (NOTE: optional)
+    """
+    import wfl.generate.buildcell
+
+    with open(buildcell_input) as bc_f:
+        buildcell_input_txt = bc_f.read()
+
+    wfl.generate.buildcell.run(
+        outputs=outputs,
+        config_is=range(n_configs),
```
bernstei, line 59: I'm pretty sure that, because the actual autoparallelized version is created by `autoparallelize()`, the input needs to be called `inputs`, not `config_is`.  

This brings the possibility that the `autoparallelize` decorator should take an optional argument with the name of the input variable, rather than hard-wiring it to `inputs`.  Is that worth doing?

wfl/cli/commands/select.py
```diff
@@ -0,0 +1,50 @@
...
+import wfl.select.by_descriptor
+from .descriptor import calculate_descriptor
+from wfl.select.simple import by_bool_func
+
+@click.command("cur")
+@click.option("--n-configs", "-N", type=click.INT, required=True,
+              help="number of configs to select")
+@click.option("--keep_descriptor", is_flag=True, help="keep the descriptor value in the final config file")
+@click.option("--kernel_exponent", type=click.FLOAT, help="exponent of dot-product for kernel")
+@click.option("--deterministic", is_flag=True, help="use deterministic (not stochastic) CUR selection")
```
bernstei, line 15: Do we want to add support for setting a seed for the stochastic one?

wfl/cli/commands/select.py
```diff
@@ -0,0 +1,50 @@
...
+@click.command("lambda")
+@click.option("--exec-code", "-e", required=True, 
+help='String to be evaluated by the lambda function. Will be substituted into `eval(\"lambda atoms: \" + exec_code)`.')
+@click.pass_context
+@opt.inputs
+@opt.outputs
+def by_lambda(ctx, inputs, outputs, exec_code):
+    """selects atoms based on a lambda function"""
+
+    if outputs.done():
```
bernstei, line 41: isn't this already implemented in `wfl.select.simple.by_bool_function`?

docs/source/index.rst
gelzinyte: > `main functions of -> main function of`

changed, thanks

> Given the generic nature of the word "workflow", what do you think about using "wfl" as the default name?

Do you think just capitalised "Workflow" would work when we're talking about the package?

tests/cli/test_descriptor.py
gelzinyte: removed

wfl/cli/cli.py
gelzinyte: just subgroups, removed it from here.

docs/source/index.rst
bernstei: Not really. Maybe all caps.  The capitalization doesn't stand out enough to me.

wfl/cli/commands/generate.py
gelzinyte: Changed to `inputs`. 

> This brings the possibility that the autoparallelize decorator should take an optional argument with the name of the input variable, rather than hard-wiring it to inputs. Is that worth doing?

Not sure if it's worth adding the complexity - any reason calling for "inputs" would be inconvenient? Also, if `inputs` aren't found, then the first argument is taken as the iterable in which case it can be called anything.

wfl/cli/commands/select.py
gelzinyte: added.

wfl/cli/commands/generate.py
bernstei: I was thinking that it was entirely ignored, so it wasn't really an input in the same way, but that's not even true - the value is saved in `Atoms.info["buildcell_config_i"]`. I'm happy to leave it as `inputs`.

wfl/cli/cli_options.py
```diff
@@ -7,7 +7,7 @@ def _to_ConfigSet(ctx, param, value):
     return ConfigSet(value)
 
 def inputs(f):
-    """Create ConfigSet for given filename(s)"""
+    """Add a standard option for the OutputSpec outputs"""
```
bernstei, line 10: New docstrings for `inputs(f)` and `outputs(f)` are switched, I think.

docs/source/index.rst
gelzinyte: Hmm maybe "WFL" then? But I don't quite like using an abbreviation still...

wfl/cli/cli_options.py
gelzinyte: of course they are... Swapped back!

docs/source/index.rst
gelzinyte: Actually, maybe this would be better discussed during the next meeting?

bernstei:
I think I'm ready to merge.  Do you agree?

docs/source/index.rst
gelzinyte: thanks! I'll merge

PR #184, opened 2022-11-07
simple changes that are needed (or just suggested by) examples in methods paper

bernstei:
As I am writing the examples in the methods paper, some obvious (often nearly trivial) changes suggest themselves to me.  These will be pushed in this PR.
 - support arbitrary kwargs for `ase.io.read` and `write` via new arguments `ConfigSet(..., read_kwargs)` and `OutputSpec(..., write_kwargs)`
 - add `wfl.map` that applies an arbitrary function in an `autoparallelize`-decorated loop
 - support for passing `AutoparaInfo` as a dict of constructor kwargs (like already supported for `RemoteInfo` and `Resources`), to reduce the number of `import`s necessary.
 - parse `WFL_EXPYRE_INFO` with `yaml.safe_load` rather than `json.load`


gelzinyte:
I think this looks good - should we merge?

Issue #143, opened 2022-08-30
Script examples that can be tested and automatically made into documentation


gelzinyte:
FInd a python equivalent of [Julia's Literate Package](https://juliapackages.com/p/literate) which would test an annotated python example script and also turn into an html-able documentation page. This way we'd be able to make sure that all our examples are up to date.

PR #185, opened 2022-11-11
Doc example dimer structures

simonwengert:
Addresses one of the tasks defined in Issue #143.


bernstei:
What do you think about moving the `assert` that's required for `pytest` to a separate code block, so as not to distract from the main example?  The automated `pytest` will have to run both, but in general I think it should run all code blocks, since some complex examples may want to break up the text that way.

simonwengert:
It's even possible to then hide this assert-codeblock completely by adding `"nbsphinx": "hidden",` to the meta data in the ipynb. In this way we can display our examples as we had before and can still use the assert-codeblocks for our pytests.

bernstei:
Could you hide the test `assert` cell in the buildcell example/test as well?

bernstei:
> I open the ipynb file via `vim` and added that line manually.

Do we need that documented someplace?  Is there a "guide to making examples", or if not, should there be one?

PR #190, opened 2022-11-22
Support for MOPAC semi-empirical FileIO calculator

gelzinyte:


bernstei:
Looks like it's failing with a `PropertyNotImplementedError` - is the ASE version that's getting installed new enough to include the bugfix that's mentioned in the MOPAC test?

gelzinyte:
> Generally looks good by eye, although I haven't looked at the failing tests. Do we want to add call to the cleanup code to the other file-io calculator wrappers?

I have moved the cleanup to wfl's FileIOCalculator so that other calculators have access to it. 

>  Looks like it's failing with a PropertyNotImplementedError - is the ASE version that's getting installed new enough to include the bugfix that's mentioned in the MOPAC test?

That's right, the pip-installed ASE doesn't have the bug fixed it seems. I can disable these tests for older ASE versions or install from gitlab for the github tests?

bernstei:
> I have moved the cleanup to wfl's FileIOCalculator so that other calculators have access to it.

That makes sense.  I'm asking whether we should add it to those other calculators as part of this PR?

> That's right, the pip-installed ASE doesn't have the bug fixed it seems. I can disable these tests for older ASE versions or install from gitlab for the github tests?

Hmmm.  I think we might as well test on the latest github (not sure which branch - `main`?  Is there a `devel`?). Most people will probably use the pypi version of ASE, but that can get pretty far behind with their tagged release schedule, and I don't recall there being anything in the main branch that didn't eventually make it into the release, so the worst case is that someone will download `wfl` expecting it to work with pypi ASE as is, but really it won't be ready until the next ASE release.

bernstei:
> So what do you mean by "add it to those other calculators"?

Go back to all the previously wrapped calculators and have them all call this cleanup routine.

> Ok, I will change github tests to use `main` of gitlab repo. `main` already has these fixes (and that's what I have been using anyway).

Sounds good.

gelzinyte:
This is a user-called function to clean up the directory `workdir_root` (if empty) in which all the individual calculation directories are stored. The individual directories get cleaned already. But the `workdir_root` cannot get removed automatically after each calculation because it might be about to be used by a parallel process.

bernstei:
Are we ready to merge?

Issue #188, opened 2022-11-16
docs: guide to writing example


bernstei:
Write a guide to writing examples (someplace in the main online docs), making them jupyter notebooks that can be automatically testing with pytest, and editing the ipynb to hide the actual final pytest testing `assert`.  See #185 for an example of how this was done.

PR #191, opened 2022-11-24
Fixes Issue #188 by providing a guide to writing examples

simonwengert:


bernstei:
Generally looks good. Do we want to say explicitly that the examples (at least ones that can be auto-tested) need to be standalone, i.e. not dependent on any files that aren't being provided? 

On a related note, is it obvious how to include a file with an example? I guess it can be added to the examples directory, but the jupyter notebook has to be careful to deal with the path correctly.  Might be a good idea to document that as well.

simonwengert:
> Generally looks good. Do we want to say explicitly that the examples (at least ones that can be auto-tested) need to be standalone, i.e. not dependent on any files that aren't being provided?

I think its not necessary to state that explicitly. I might be wrong, but I think it should be obvious that all required information needs to be provided. I would rather add it later if it turns out we actually need it. Also, in case someone tries to push such an example to the repo, the CI test suite will fail and tell the contributor that he/she did something wrong.

> On a related note, is it obvious how to include a file with an example? I guess it can be added to the examples directory, but the jupyter notebook has to be careful to deal with the path correctly. Might be a good idea to document that as well.

Never tried it, but would guess they can be added to the examples directory, too. Do we want to have all such additional files in one common directory? Or should we distribute them into different (example-specific) sub-directories to avoid any conflict that might appear from having to deal with multiple examples.

bernstei:
> I think its not necessary to state that explicitly. I might be wrong, but I think it should be obvious that all required information needs to be provided. I would rather add it later if it turns out we actually need it. Also, in case someone tries to push such an example to the repo, the CI test suite will fail and tell the contributor that he/she did something wrong. 

OK

> Never tried it, but would guess they can be added to the examples directory, too. Do we want to have all such additional files in one common directory? Or should we distribute them into different (example-specific) sub-directories to avoid any conflict that might appear from having to deal with multiple examples.

I don't particularly like a single common directory for the files from all examples.  I think each example's files should live close to its own script, but I'm happy to leave the people dealing with any given example the decision on whether it should just be a file in the same directory or in a deeper subdirectory. That might depend on whether it's one file or many, for example.

bernstei:
I'm happy to merge this.  @simonwengert @gelzinyte anything else people want to change before that?

gelzinyte:
Looks great, thanks!

PR #192, opened 2022-12-01
CLI for error table & scatterplot

gelzinyte:


wfl/cli/commands/error.py
```diff
@@ -1,6 +1,89 @@
...
+def print_df(df, precision=3):
+
+    pd.set_option('display.max_rows', None)
+    pd.set_option('display.max_columns', None)
+    pd.set_option('display.colheader_justify', 'center')
+    pd.set_option('display.precision', precision)
+    print(df)
+
+
+def format_errors(errors):
```
bernstei, line 79: Might be nice to have this and `print_df` available from python, not just command line. Would just need to move them into `fit/error.py` and call those routines from here.  It might also be nice to add an optional file argument to `print_df`, defaulting to `sys.stdout`. Or we could have a version that returns a formatted string with `io.StringIO`.

Also, `pd.set_option` is a global setting, which wouldn't be so friendly if it's in a function, not the command line, so we'd need to figure out how to get pandas to apply the options only to this table output operation.   I think you can use `DataFrame.to_string()` with some optional arguments that take the place of these options.

bernstei: Yes - I think the optimal thing is to have `format_errors()` return a string, probably generated by `df.to_string(..., max_rows=...,...)` with other optional args, and move that function to `fit/error.py`.  Then the command line can just print that string, and other people can call the function as well and print it wherever they want.

gelzinyte: > Might be nice to have this and print_df available from python, not just command line. 

> It might also be nice to add an optional file argument to `print_df`, defaulting to `sys.stdout`. 

sounds good. 

> `pd.set_option` is a global setting

Ah of course. I was going to fix that with a `with` scope, but I guess I can convert to correctly formatted string before printing. I would rather `format_errors()` returned a dataframe though, because then other people have options to convert it to markdown, html, latex, ... not only pre-formated string.

bernstei:
I'm also getting the following error when I try to run it on some real data
```
tin 1973 : wfl error -cpp ace_ -rpp REF_ -i ACE_fit_params.3_12_weight_from_sigma_E_0.2_brr.fitting_database.extxyz 
Traceback (most recent call last):
  File "/home/Software/python/system/extra/lib64/python3.9/site-packages/pandas/core/indexes/base.py", line 3629, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 136, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 163, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 5198, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 5206, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'MAE'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/cluster2/bernstei/.local/bin/wfl", line 33, in <module>
    sys.exit(load_entry_point('wfl', 'console_scripts', 'wfl')())
  File "/home/cluster2/bernstei/.local/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/cluster2/bernstei/.local/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/cluster2/bernstei/.local/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/cluster2/bernstei/.local/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/cluster2/bernstei/.local/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/cluster2/bernstei/.local/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/cluster2/bernstei/src/work/workflow/wfl/cli/commands/error.py", line 53, in show_error
    df = format_errors(errors)
  File "/home/cluster2/bernstei/src/work/workflow/wfl/cli/commands/error.py", line 85, in format_errors
    df["MAE"] = df["MAE"].apply(lambda x: x*1e3)
  File "/home/Software/python/system/extra/lib64/python3.9/site-packages/pandas/core/frame.py", line 3505, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/home/Software/python/system/extra/lib64/python3.9/site-packages/pandas/core/indexes/base.py", line 3631, in get_loc
    raise KeyError(key) from err
KeyError: 'MAE'
```

bernstei:
The problem appears to be that I get an empty errors structure returned from the error calculator.

bernstei:
Aha - the CLI passes empty lists of property names rather than `None` if you don't specify any config or atom properties on the command line, but the error calculator itself only does the default behavior if those are passed as `None`.

bernstei:
```
diff --git a/wfl/fit/error.py b/wfl/fit/error.py
index 38f3732..4063295 100755
--- a/wfl/fit/error.py
+++ b/wfl/fit/error.py
@@ -45,6 +45,10 @@ def calc(inputs, calc_property_prefix, ref_property_prefix,
             each category and property, for parity plots
     """
 
+    if len(config_properties) == 0 and len(atom_properties) == 0:
+        # trigger default behavior
+        config_properties = None
+        atom_properties = None
     # default properties
     if config_properties is None and atom_properties is None:
         config_properties = ["energy/atom", "virial/atom/comp"]
```

bernstei:
Also, some sorting of the categories could be helpful.  This is definitely not what we want.
```
                                RMSE       MAE    count  units 
energy/atom      config_0     3.19e+00  3.19e+00      1  meV/at
                 _ALL_        1.51e+02  2.38e+01   3522  meV/at
                 config_1     3.78e+00  3.78e+00      1  meV/at
                 config_2     3.11e+00  3.11e+00      1  meV/at
```

wfl/cli/commands/error.py
gelzinyte: Added this in [e8d082b](https://github.com/libAtoms/workflow/pull/192/commits/e8d082be11a733f1ec871dc78782bb8b2988f966)

gelzinyte:
> Also, some sorting of the categories could be helpful. This is definitely not what we want.

Ok, now sorting alphabetically with `_ALL_` at the end. Or should we do something else?

wfl/cli/commands/error.py
```diff
@@ -1,6 +1,70 @@
...
 
 @click.command("error")
-def show():
-    print("not yet implemented")
-    pass
\ No newline at end of file
+@click.option('--calc-property-prefix', '-cpp', required=True, 
+    help="prefix for calculated (predicted) properties (energy, forces, ...)")
+@click.option("--ref-property-prefix", '-rpp', required=True, 
+    help="prefix for properties taken as reference (e.g. from electronic strucutre codes)")
```
bernstei, line 13: typo: `strucutre` -> `structure`

wfl/fit/error.py
```diff
@@ -375,9 +378,56 @@ def _annotate_parity_plot(ax, property, ref_property_prefix, calc_property_prefi
...
+    file: obj, default sys.stdout
+        Object to write the table to. 
+    """
+
+    df = errors_to_dataframe(errors) 
+    with pd.option_context('display.max_rows', None,
+                           'display.max_columns', None,
+                           'display.precision', precision,
+                           'display.colheader_justify', 'center'):
+        print(df, file=sys.stdout)
```
bernstei, line 405: By analogy to things like `json.dumps()`, this should return a string, not actually print. Why not use `DataFrame.to_string()`?

wfl/cli/commands/error.py
bernstei: I'd still prefer to have a routine that returns a string.  See my comment below.

bernstei:
> The problem appears to be that I get an empty errors structure returned from the error calculator.

Should we test for this and report a clear error, rather than letting it fail on the `'MAE'` key lookup?

bernstei:
> Aha - the CLI passes empty lists of property names rather than `None` if you don't specify any config or atom properties on the command line, but the error calculator itself only does the default behavior if those are passed as `None`.

A test is failing because we're checking on the lengths but the test is passing `None`. Maybe we should drop the newly added if block, and replace the next one with
```
    if ((config_properties is None or len(config_properties) == 0) and
        (atom_properties is None or len(atom_properties) == 0)):
```

bernstei:
BTW, not a part of this PR conceptually, but I've been testing the error command line together the the `wfl eval gap` command line, and I think that the param filename option really needs a short version, since it's always required.

bernstei:
We need a check for an empty errors dict. That can happen if you, e.g., specify the wrong prefix for either calculation, and then you get mysterious looking pandas errors (e.g. complaining about the `_ALL_` column being missing, in my case).

wfl/cli/commands/error.py
```diff
@@ -1,6 +1,70 @@
...
+    errors, diffs, parity = ref_err_calc(
+        inputs=inputs, 
+        calc_property_prefix=calc_property_prefix,
+        ref_property_prefix=ref_property_prefix,
+        config_properties=config_properties,
+        atom_properties=atom_properties,
+        category_keys=category_keys,
+        weight_property=weight_property)
+    
+    print(errors)
```
bernstei, line 53: I think this unformatted print statement should be gone, right?

bernstei:
After using the command line interface I decided that I really want to be able to generate a table with one column per property, and one row per category (and perhaps file).  It's much better, in my opinion, for comparing potentials. 

I think that should be an option. I can think about how to implement it cleanly.

wfl/fit/error.py
gelzinyte: I see what you meant, please resolve this if the update looks ok.

gelzinyte:
> The problem appears to be that I get an empty errors structure returned from the error calculator.

> Should we test for this and report a clear error, rather than letting it fail on the 'MAE' key lookup?

Why was the dictionary empty? I would think it should always be populated, so that would be a bug in `wfl.fit.error.calc`?

gelzinyte:
> BTW, not a part of this PR conceptually, but I've been testing the error command line together the the wfl eval gap command line, and I think that the param filename option really needs a short version, since it's always required.

Should we try to be consistent and say that short versions are initials of long versions? At least for properties that go with multiple commands? Namely: 

`--prop-prefix`  (currently `-p`) -> `-pp`
`--param-fname` -> `-pf`
`--extra-info` (currently `-i`) -> `-ei`

bernstei:
> > After using the command line interface I decided that I really want to be able to generate a table with one column per property, and one row per category (and perhaps file). It's much better, in my opinion, for comparing potentials.
> > I think that should be an option. I can think about how to implement it cleanly.
> 
> I'd be up for this - I rarely want to see both MAE and RMSE. I added a possible change. Not sure how to handle different files, though. Could do double-column-label (property with multiple file sub-columns)? It might be useful to see count too - but that would mean count of structures, not instances of given property.

For my own use I really want the columns to be properties, and everything else (filename and category) to be rows, essentially regarding the file as another part of the category. I can try to implement that.

bernstei:
> > BTW, not a part of this PR conceptually, but I've been testing the error command line together the the wfl eval gap command line, and I think that the param filename option really needs a short version, since it's always required.
> 
> Should we try to be consistent and say that short versions are initials of long versions? At least for properties that go with multiple commands? Namely:
> 
> `--prop-prefix` (currently `-p`) -> `-pp` `--param-fname` -> `-pf` `--extra-info` (currently `-i`) -> `-ei`

That'd be good, yes.

bernstei:
Looks generally good. I'll deal with the properties -> columns, files + categories -> rows. Should I make it the only output?  The default output?

Can you do the flag short names, which I think is the only other thing left?

tests/cli/test_error.py
```diff
@@ -0,0 +1,33 @@
...
+              f"--fig-name {fig_name} "
+
+
+    print(command)
+
+    runner = CliRunner()
+    result = runner.invoke(cli, command)
+    assert result.exit_code == 0
+
+    assert False
```
bernstei, line 33: Is this some relic of debugging?

gelzinyte:
> Looks generally good. I'll deal with the properties -> columns, files + categories -> rows. Should I make it the only output? The default output?

Yes I think that sounds good. 


> Can you do the flag short names, which I think is the only other thing left?

Done

gelzinyte:
Thanks. It might have been a typo in one of the cli options - fixed that in the latest commit.

bernstei:
Do you think we need to support reporting both RMSE and MAE anywhere except in the original dict? Or can we make all the formatting/plotting routines require that the user specify one or the other?

gelzinyte:
I think we can default to one of them

bernstei:
@gelzinyte what do you think of the current table format?  Whenever that's settled, I'm ready to merge.

bernstei:
> Looks great! There's a `print("BOB", df.columns)` left behind, but can merge apart from that.

Now you know my secret debugging label.

bernstei:
One thing I was considering was to add to the function that calls `to_string` the ability to pass arbitrary kwargs down to that function (e.g. if you do or don't want `sparsify` or `justify`.  Do you think that's useful enough to be worth it?

gelzinyte:
I think defaults would be enough, but having a way to give kwargs won't interfere

PR #197, opened 2022-12-02
Add optional flush argument to OutputSpec() constructor

bernstei:


bernstei:
Nearly trivial, but I guess we should decide if we should make `flush=True` the new default.  It is safer, and I don't think it's likely to be detectably slower, although in principle it could be.  But I doubt anyone would parallelize over an operation so fast that flushing output would add anything significant.

gelzinyte:
I think `flush=True` is fine for a default, so will merge.

PR #201, opened 2022-12-06
Rewrite Castep wrapper 

gelzinyte:


wfl/calculators/wfl_fileio_calculator.py
```diff
@@ -66,21 +66,34 @@ def setup_rundir(self):
...
-            if list(Path(self.directory).iterdir()) == []:
-                warnings.warn(f"scratchdir {self.directory} is not empty, not deleting.")
+            if list(Path(directory).iterdir()) == []:
+                warnings.warn(f"scratchdir {directory} is not empty, not deleting.")
             else:
-                Path(self.directory).rmdir()
-                self.directory = '.' 
+                Path(directory).rmdir()
+                if self.__name__ == "Castep":
+                    self._direcotyr = '.'
```
bernstei, line 94: Typo.  Also, I see that this bit of code will also have to remain Castep-specific.

wfl/calculators/wfl_fileio_calculator.py
```diff
@@ -66,21 +66,34 @@ def setup_rundir(self):
...
+            directory = self.directory
+            
+        utils_clean_rundir(directory, self._wfl_keep_files, _default_keep_files, calculation_succeeded)
         if self._wfl_scratchdir is not None:
-            for f in Path(self.directory).glob("*"):
+            for f in Path(directory).glob("*"):
                 shutil.move(f, self._cur_rundir)
-            if list(Path(self.directory).iterdir()) == []:
-                warnings.warn(f"scratchdir {self.directory} is not empty, not deleting.")
+            if list(Path(directory).iterdir()) == []:
```
bernstei, line 89: Is this test actually correct?  It seems to contradict the warning on the next line that it is _not_ empty.

wfl/calculators/wfl_fileio_calculator.py
bernstei: Also, how important is it that we reset `self.directory`? Won't it get re-set next time, but `setup_rundir`?

wfl/calculators/wfl_fileio_calculator.py
gelzinyte: Aha, this seems to be incorrect, changed this to `!=`.

wfl/calculators/wfl_fileio_calculator.py
gelzinyte: Yes, I think we can not reset `self.directory` (I don't think resetting it to `"."`  was correct anyway)

wfl/calculators/wfl_fileio_calculator.py
```diff
@@ -66,21 +66,34 @@ def setup_rundir(self):
             dir_name = str(self._cur_rundir.resolve()).replace("/", "", 1).replace("/", "_")
             directory = self._wfl_scratchdir / dir_name
             directory.mkdir(parents=True, exist_ok=True)
-            self.directory = directory
         else:
-            self.directory = self._cur_rundir
+            directory = self._cur_rundir
 
+        if self.__name__ == "Castep":
```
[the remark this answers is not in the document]
gelzinyte: That's better - I've made wfl Castep have a `self.directory` and set `sefl._directory` based on that, but now in the castep wrapper.

wfl/calculators/wfl_fileio_calculator.py
bernstei: Thanks.  Not sure it's worth doing as part of this PR, but we should probably have a pytest that would have caught this mistake.

wfl/calculators/castep.py
```diff
@@ -1,133 +1,131 @@
...
+        finally:
+            # ASE castep calculator does not
             # always (ever?) raise an exception when it fails.  Instead, you get things like 
             # stress being None, which lead to TypeError when save_results calls get_stress().
-            try:
-                save_results(at, properties_use, output_prefix)
-            except TypeError:
-                calculation_succeeded=False
+            for property in properties:
+                result = self.get_property(property)
```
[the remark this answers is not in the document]
bernstei: OK, makes sense.

wfl/calculators/castep.py
```diff
@@ -1,133 +1,134 @@
...
+
+        if kwargs.get("castep_pp_path", None) is None:
+            # make sure we are looking for pspot if path given
+            kwargs["find_pspots"] = True
+
+        # WFLFileIOCalculator is a mixin, will call remaining superclass constructors for us
+        super().__init__(keep_files=keep_files, rundir_prefix=rundir_prefix,
+                         workdir=workdir, scratchdir=scratchdir, **kwargs)
+
+        self.directory = self._directory
```
bernstei, line 80: I think it's worth commenting where stock Castep is non-standard.  Can we add something like 
```
# Copy from Castep's `_directory` attr to more standard `directory` attr so the rest of `wfl` can find it
```
I'll try to find the other places where I think such comments are worth it.

wfl/calculators/castep.py
```diff
@@ -1,133 +1,134 @@
...
+
+        if atoms is not None:
+            self.atoms = atoms.copy()
+
+        # this may modify self.parameters, will reset them back to initial after calculation
+        nonperiodic, properties = self.setup_calc_params(properties)
+
+        # from WFLFileIOCalculator
+        self.setup_rundir()
+        self._directory = self.director
```
bernstei, line 97: typo, also comment about Castep non-standard behavior

wfl/calculators/wfl_fileio_calculator.py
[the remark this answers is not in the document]
bernstei: thanks.

gelzinyte:
So after all ASE Castep wrapper has setter and getter methods overwritten and they check that only the allowed attributes are set unless they start with an underscore. So I think the easiest way forward is to set `self._directory` in `WFLFileIOCalculator` for a Castep calculator...

bernstei:
> So after all ASE Castep wrapper has setter and getter methods overwritten and they check that only the allowed attributes are set unless they start with an underscore. So I think the easiest way forward is to set `self._directory` in `WFLFileIOCalculator` for a Castep calculator...

OK.  Let's just see if we can restrict the Castep-specific stuff to as few places as possible.

PR #203, opened 2022-12-12
pytest for wfl command line examples

bernstei:
Update `test_doc_examples.py` to correctly handle (at least specially)crafted `wfl` command line tests. 

Note that this initial commit does not include the files needed for the example test.

### DIFF
--- a/docs/source/examples.cli.ipynb
+++ b/docs/source/examples.cli.ipynb
@@ -1,0 +1,91 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "72125898",
+   "metadata": {},
+   "source": [
+    "# command line examples"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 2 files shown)


bernstei:
> I'm going to try to convert the relative paths to absolute paths starting from the location of `wfl.__file__`, I think.

So I figured out how to do this, but I don't think that for a notebook that's the right approach - I think it makes more sense for the notebook to use _its own_ directory's examples, rather than the one that's linked to the wfl version that `import wfl` returns. Actually, if `wfl` is installed from pypi rather that a local repo clone, there may not even _be_ an examples directory on the system.

bernstei:
Explanation: I put in a framework for pytest to automatically test docs examples that use the command line. However, I was unable to put in a complete example.  The framework is in `docs/source/examples.cli.ipynb`.  Someone else needs to upload the corresponding files to `docs/example_files` so that command will run and test will pass.

PR #205, opened 2022-12-19
Fhiaims calculator

simonwengert:
Implemented a calculator for `FHI-Aims` compatible with the `generic` calculator.

Comprises the Calculator Class, tests for it and an update in the documentation.


simonwengert:
Hm, on my fork I am passing the test suit (https://github.com/simonwengert/workflow/actions/runs/3732830524). I'm wondering why `tests/test_remote_run.py::test_resubmit_killed_jobs FAILED` here.

bernstei:
I just forced a rerun, in case it was some weird transient issue.

bernstei:
It's looking like a genuine problem in something related to wfl and expyre, but it has nothing to do with fhi-aims, so I have no idea why it's failing now.   I'll try to fix it properly.

bernstei:
I believe this was caused by a subtle bug in ExPyRe, interacting with how exactly the queuing system behaves when a job is killed when it's out of walltime.  I believe that https://github.com/libAtoms/ExPyRe/pull/29 should fix it, but I want to make sure that expyre patch is OK before I merge it and rerun the tests here.

simonwengert:
Thanks for fixing the issue!

This PR contains some tests for the FHI-Aims calculator and large parts of it require an installation of FHI-Aims. Compiling it as part of the CI is problematic, right? (Since FHI-Aims requires a license or reasons related to this)

bernstei:
Yes, I'm not aware of any legal way to do such tests on the CI, which is why we don't actually run the VASP or CASTEP tests in the CI either.  For VASP I just get in the habit of running them locally whenever there's a PR that seems like it might affect them, and hopefully whoever did CASTEP and now you for FHI-Aims can do it for those tests.

tests/calculators/test_aims.py
```diff
@@ -0,0 +1,166 @@
...
+def test_setup_calc_params(parameters_nonperiodic):
+
+    parameters = parameters_nonperiodic
+    parameters_periodic = {
+        'k_grid': '1 1 1',
+        'k_grid_density': 0.1,
+        'k_offset': 0.1,
+        'relax_unit_cell': 'full',
+        'external_pressure': 10,
+        'compute_analytical_stress': '.true.',
```
bernstei, line 52: Having settings like `compute_analytical_stress` required to be set correctly outside the calculator makes it impossible to cleanly apply the workflow to a combination of periodic and non-periodic configurations, e.g. what you might end up with when you are reevaluating an entire database which includes crystal cells as well as isolated atoms. 

Could we have this dealt with automatically inside the calculator, which should know both whether a configuration is periodic (`ase.Atoms.pbc`) and whether a stress has been requested from that calculation (if it's even slow enough to be worth turning off)

wfl/calculators/aims.py
```diff
@@ -0,0 +1,165 @@
...
+    workdir: str / Path, default . at calculate time
+        Path in which the run directory will be created.
+    scratchdir: str / Path, default None
+        Temporary directory to execute calculations in and delete or copy back results (set by ```keep_files```)
+        if needed.  For example, directory on a local disk with fast file I/O.
+    calculator_exec: str
+        command for Aims, without any prefix or redirection set.
+        For example: "srun -n 4 /path/to/aims.*.scalapack.mpi.x".
+        Mutually exclusive with ```command```.
+    get_k_grid: func(Atoms) -> str='<k1> <k2> <k3>', default None
```
bernstei, line 50: Is there really a need for a general function here, or is supporting something equivalent to k-spacing (which VASP and CASTEP use) sufficient? If there is a need for a general function, that almost certainly applies to other calculators, and we should provide a general-use function.

tests/calculators/test_aims.py
simonwengert: The current version handles this automatically inside the calculator in some sense. For each structure it detects if its periodic or non-periodic. If its periodic it starts the calculation with all the (FHI-Aims) settings passed. If its non-periodic, it removes all of the additional key-value pairs that need to be passed for a periodic calculation.

wfl/calculators/aims.py
simonwengert: There is an argument `k_grid_density` in FHI-Aims which might be what we would like to have. I personally never used it (always used `k_grid`). I'm trying to find out more about it. If `k_grid_density` is the FHI-Aims equivalent to the k-spacing arguments in other codes (and I think it is) I can also remove the `get_k_grid`-functionality.

tests/calculators/test_aims.py
bernstei: OK - I guess if the workflow removes invalid keywords it's OK.  Maybe we should document in the calculator's docstring what keywords get removed when pbc is False.

simonwengert: Agreed. I'll adjust the docstring.

Related to this: What is your opinion on this alternative that will allow the user specifying individual settings for each configuration in a database?
`generic.run()` gets an additional argument `calculator_settings` which is a list of dictionaries (each dictionary representing the `kwargs` used to define the Calculator in the form `Calculator = (initializer, args, kwargs)`, e.g. one `kwargs`-setting for PBC, one for non-PBC, one for isolated atoms or so). When looping over the configurations in the users database, the calculator gets constructed individually for each config (i.e. shifting the `construct_calculator_picklesafe()` inside the loop) where the required `kwargs` gets selected from the list in `calculator_settings` based on an index that needs to be specified in the Atoms.info for each config (something like `Atoms.info['calculator_settings_index'] = 1`).

bernstei: I thought about this before, although not with a short list and an index, but zipping the configs and a list of calculators, one per config.

What exactly is the use case? If it's just periodic vs. non-periodic, I think I prefer that being handled automatically in the calculator, like we've done so far. Do you have a specific example where that can't be done cleanly, or another use for this feature? 

The only one I can thin think of is to be able to explicitly specify k-point meshes without needing to rely on k-spacing, which can have problems with small strain causing a discontinuous jump in the precise number along particular directions.  It's pretty unusual, though, and yet still perhaps sufficiently ubiquitous across calculators that I' might prefer a special `Atoms.info["DFT_k_pt_mesh"]` field which all the wfl DFT calculators know about, so users don't have to change anything when changing DFT calculators.

simonwengert: > What exactly is the use case? If it's just periodic vs. non-periodic, I think I prefer that being handled automatically in the calculator, like we've done so far. Do you have a specific example where that can't be done cleanly, or another use for this feature?

The first thing I was thinking about was that the isolated atoms can sometimes be tricky and need some more specific setting. Other situations might be that a user wants to set a keyword such that the geometry gets optimized, but only for a certain part of the structures in the database, or a user is interested in differences (e.g. different VdW-corrections or so) for the structures in a database.

bernstei: I think if it's just two groups (isolated vs. everything else), I think it's not worth the complexity - you can just do two separate evaluation calls, and the isolated atoms are usually pretty cheap anyway.

I'm also down on using the DFT calculator to do the relaxation, although others may like that.  I prefer to use our minimizers, whose behavior I understand, with the calculator just providing gradients.  I do admit it's typically less efficient, though.

wfl/calculators/aims.py
simonwengert: Ok, so `k_grid_density` is the keyword we can use here instead of introducing the `get_k_grid`-functionality. I'll remove it therefore.
The bad news is that the latest ASE release (3.22.1) does not support this keyword in the Aims-Calculator. (Version 3.23.0b1 does, though.)

tests/calculators/test_aims.py
gelzinyte: I would also lean towards less complexity unless there's clear demand for the functionality or significant improvement. 

As for DFT relaxations, it was 10 times more efficient to do it directly with ORCA, so I explicitly use that. Wfl ORCA wrapper has that implemented.

gelzinyte: that was number of geometry optimisation steps. ORCA is using internal coordinates and I would imagine it can reuse some of the electronic structure information from the previous geometry opt step. I wonder if internal using internal coordinates is a bit less beneficial for materials than it is for organic molecules.

bernstei:
> The CI test suite fails already before running the pytests. There seems to be an issue when trying to install MOPAC. Any idea what causes this?

Some conda dependency issue.  I hate conda.

bernstei:
I don't see anything that's changed w.r.t. what's being done in the CI with conda.  It's only used to install pip and then mopac, and yet it still fails, and it was working a few weeks ago.  I really have no idea. @gelzinyte ? Have you had any such issues with mopac from conda?

gelzinyte:
@simonwengert merging the current `main` branch into yours should resolve the mopac test issues.

simonwengert:
:+1: Works now!

PR #212, opened 2023-01-23
Update gap_rss_iter_fit.py for recent changes to routines it uses

bernstei:
`buildcell.create_output` now returns string
`fit.error.calc` only computes errors, without evaluating with potential, and returns different quantities


bernstei:
For now waiting on the gap rss iter fit test, which is running locally.

PR #222, opened 2023-01-31
daisy-chained example

gelzinyte:
This is (an unpolished draft of) the long daisy-chained workflow example that involves quite a few different operations and different ways to interact with them. 

@simonwengert, @bernstei 
- what do you think in general? 
- Any parts to skip or parts to add? 
- Any better unit-tests than just checking for presence of files? 
- Originally, this showed how to set remote execution via environment variable and python. Currently I have left the in-python remote execution, but am not actually using it in the end for github-workflow purposes. Any better way to showcase this? 
- I left gap fitting in, but now I think that ACE fitting would be better, since we already have an iterative-fitting GAP example? 
- Do you think much more explanation is needed or can/should we treat this more as a copy-and-paste-code-example-pool and expect people to have a sense of what's happening from the rest of documentation? (I would prefer the less verbose later.)

@simonwengert thanks for working out how to document and test jupyter notebooks - writing this example was much easier than the old ones!


gelzinyte:
Note to myself on to do: come back to the clumsy piece of code that calculates atomization energies. I would like to be able to overwrite an existing file rather than creating new ones. Need to check if PR #210 allows to do this or if more modifications would be needed.

simonwengert:
> what do you think in general?
In general I really like this long-ish example. It's well structures and in my opinion its easy to understand what's going on and why. 

> Any parts to skip or parts to add?
We could save one part of the example by shifting the _Filter out unreasonable structures_ into the _Run Molecular Dynamics simulation_ part by directly selecting the structures during the MD run. In addition to saving one part, we then also showcase the `traj_select*`-functionality for MD runs. And we still have a representative of `wfl.select` due to the CUR sampling below.

> Any better unit-tests than just checking for presence of files?
When initializing the MD with a defined set of velocities, is the rest of the example then deterministic? If this is the case, we could check if the obtained errors at the end are as expected.

> I left gap fitting in, but now I think that ACE fitting would be better, since we already have an iterative-fitting GAP example?
Indeed, having an ACE example would be nice. Either with this example or a separate one. Since it's a long example and we don't want to lose people on the way GAP might be a preferred choice. I guess there are more people that have some experience with GAP then with ACE.

gelzinyte:
> We could save one part of the example by shifting the Filter out unreasonable structures into the Run Molecular Dynamics simulation part by directly selecting the structures during the MD run. In addition to saving one part, we then also showcase the traj_select*-functionality for MD runs. And we still have a representative of wfl.select due to the CUR sampling below.

I like this, will add!

> When initializing the MD with a defined set of velocities, is the rest of the example then deterministic? If this is the case, we could check if the obtained errors at the end are as expected.

That would mean loading the trajectory start from file, which I think doesn't (by default) store enough significant digits. Also quite like that the tutorial doesn't depend on any files. But I agree, it would be nice to check the errors - maybe add a very loose assert on the RMSEs?

bernstei:
Why not just set the random seed?

wfl/fit/error.py
```diff
@@ -379,10 +379,16 @@ def select_units(prop, plt_type):
         prop = re.sub(r"/comp\b", "", prop)
     if "forces" in prop:
         prop = "forces"
-    try:
-        return units_dict[prop][plt_type]
-    except KeyError:
-        return "unknown units"
+
+    if "energy/atom" in prop:
```
bernstei, line 383: What's the specific use case for allowing arbitrary properties that happen to include things like `"energy/atom"` ?

bernstei: Is it other `"/<something"` suffixes?

bernstei: If so, might be nice to put in a comment to explain that.

wfl/utils/configs.py
```diff
@@ -69,7 +69,7 @@ def atomization_energy(inputs, outputs, prop_prefix, property="energy", isolated
         "default" matches "isolated_atom" or "IsolatedAtom".
     """
 
-    if outputs.is_done():
+    if outputs.done():
```
bernstei, line 72: did we change this `all_written()` now?

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+    "}\n",
+    "\n",
+    "# set to None for github testing purposes\n",
+    "remote_info=None\n",
+    "\n",
+    "inputs = md.sample(\n",
+    "    inputs=inputs, \n",
+    "    outputs=outputs,\n",
+    "    calculator=xtb_calc,\n",
+    "    autopara_info = AutoparaInfo(\n",
```
bernstei, line 250: I intended to code these up so that you could pass a dict of constructor kwargs and not have to actually import `AutoparaInfo`. We should think about whether that's enough of a convenience to be worth using routinely, or if we always want to import the object definition and construct it explicitly.

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Reference calculator \n",
+    "\n",
+    "To parallelise (non-file-based) calculators, they need to be given to workflow functions as \n",
```
bernstei, line 95: Technically this is only true for more complex ones, which can't be pickled (like quippy Potential).  Do we want to get into that level of detail here?  At least I'd change it to "the safest way to specify ... is to ..."

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+    "\n",
+    "inputs = generic.run(\n",
+    "    inputs=inputs,\n",
+    "    outputs=outputs,\n",
+    "    calculator=xtb_calc,\n",
+    "    properties=[\"energy\"],\n",
+    "    output_prefix=\"xtb_\"\n",
+    ")\n",
+    "\n",
+    "outputs = OutputSpec(isolated_at_fname)\n",
```
bernstei, line 144: I'm not sure that we should be encouraging users to deal with this kind of OutputSpec.store() loop routinely.  Honestly, what _I_ would probably like is acknowledge the fact that the calculation result is in memory, modify the Atoms objects in memory, and then write them to a file.  Would it be helpful to have a `ConfigSet.write(filename(s))` method?

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+    "    'C1=CCC(Cc2ccc(CC3CC3)cc2)C1', \n",
+    "    'C1=CC(c2ccccc2)=CCC1', \n",
+    "    'C/C=C/CCCC[C@H](C)C(C)(C)C', \n",
+    "    'C=C[C@@H]1C/C=C/CCCCCCCC1', \n",
+    "    'C[C@H](CC(C)(C)C)[C@@H](C)C(C)(C)C', \n",
+    "    'CC/C=C\\\\C[C@@H](C)c1cccc(C)c1C', \n",
+    "    'C=C1CC2c3ccccc3C1c1ccccc12']\n",
+    "\n",
+    "outputs = OutputSpec(\"1.ch.rdkit.xyz\")\n",
+    "inputs = smiles.run(all_smiles, outputs=outputs)"
```
bernstei, line 185: I would encourage less generic names than always using `inputs` and `outputs`.  Especially when you get to these more complex parts of the workflow, this looks weird, and in the next code block, where you're relying on `inputs` to contain the smiles _output_, it's extra weird.  How about something like `smiles_configs` ?

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "\n",
+    "def are_xtb_force_comps_reasonable(at):\n",
+    "    force_comps = at.arrays[\"xtb_forces\"]\n",
+    "    return np.all(np.abs(force_comps) < 8 )\n",
```
bernstei, line 275: quite minor, but isn't it more physically reasonable to consider the total force on each atom, so `np.all(np.linalg.norm(force_comps, axis=1) < 8)` ?

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tmp_outputs = OutputSpec()\n",
+    "outputs = OutputSpec(\"4.ch.rdkit.md.traj.filtered.soap.xyz\")\n",
+    "\n",
+    "# a nice feature of `ConfigSet`: \n",
+    "# can check if output from this step is already created and not re-execute it\n",
+    "if outputs.done():\n",
```
bernstei, line 304: FWIW, I personally wouldn't bother with this check in this case. I'd do the actual descriptor calculation to a file, which will be auto-skipped, and assume that the cost of the averaging operation is trivial.  Also, I thought I added a `map` function which will do fairly arbitrary operations _with autoparallelization_.

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+    "gap_fname='gap.xml'\n",
+    "\n",
+    "# process configs before fitting\n",
+    "inputs = list(inputs)\n",
+    "for at in inputs:\n",
+    "\n",
+    "    # set the cell for gap_fit\n",
+    "    at.cell = [50, 50, 50]\n",
+    "\n",
+    "    # gap_fit cannot parse the xyz correctly with the SOAP descriptor\n",
```
bernstei, line 386: Do we understand why not?

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+    "        \"partitions\" : \"any$\"}, \n",
+    "    \"partial_node\": True,\n",
+    "    \"check_interval\": 5, \n",
+    "    \"pre_cmds\" :[\"conda activate dev\"],\n",
+    "}\n",
+    "\n",
+    "# set to None for github testing purposes\n",
+    "remote_info = None\n",
+    "\n",
+    "if not Path(gap_fname).exists(): \n",
```
bernstei, line 501: Is this really not auto tested for skipping?

gelzinyte: No, but I think it should be. Have added `skip_if_present=False` to `run_gap_fit`, similarly to `run_ace_fit`. Or should it default to True?

gelzinyte:
I am backing out of changing from GAP to ACE, mainly because GAP is just a pip install, so it's less of a barrier for someone to try this tutorial. But having a separate example would be nice.

wfl/fit/error.py
bernstei: I'm not sure it's a good idea to have things that could match such arbitrary property names.  Is it too complicated to only guess the exact names of a few specific properties, and pass a dict of units (property names are keys) that can extend/override the built-in guesses?

docs/source/examples.daisy_chain_gap.ipynb
```diff
@@ -0,0 +1,746 @@
...
+    "remote_info = RemoteInfo(\n",
+    "    sys_name = \"local\",\n",
+    "    job_name = \"gap-eval\",\n",
+    "    resources = resources,\n",
+    "    partial_node = True,\n",
+    "    check_interval=10, \n",
+    "    input_files=[\"gap.xml*\"], \n",
+    "    pre_cmds = [\"conda activate dev\"])\n",
+    "\n",
+    "remote_info = None\n",
```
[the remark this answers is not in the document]
bernstei: I'm pretty sure it's there. The expyre config file is in `workflow/.github/workflows_assets/config.json`

docs/source/examples.daisy_chain_gap.ipynb
bernstei: Having that kind of example someplace, sure.  That it's needed in the example we're using for this paper I'm less convinced of.

docs/source/examples.daisy_chain_gap.ipynb
bernstei: Do you have the file? I could take a look.

docs/source/examples.daisy_chain_gap.ipynb
bernstei: I don't have strong feelings about it.  I suppose we could encourage the safer approach which is to use the classes.

docs/source/examples.daisy_chain_gap.ipynb
bernstei: Should we directly pass one or more filenames, or should we pass an outputspec, so it could be used like `ConfigSet.write(OutputSpec(filename(s)))` ?

docs/source/examples.daisy_chain_gap.ipynb
bernstei: Did I really not already have that?  Was it in the multistep GAP, but just not in simple?

docs/source/examples.daisy_chain_gap.ipynb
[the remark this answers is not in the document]
bernstei: The CI task copies it to `~/.expyre/config.json`

docs/source/examples.daisy_chain_gap.ipynb
bernstei: Or should it actually be an `OutputSpec.write(ConfigSet)` method?  You could use it like
```
OutputSpec("filename.xyz").write(configset_of_stuff)
```

docs/source/examples.daisy_chain_mlip_fitting.ipynb
```diff
@@ -0,0 +1,726 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "train_inputs = ConfigSet([list(processed_cur_selection)[0::2], read(isolated_at_fname, \":\")])\n",
+    "test_inputs = ConfigSet(list(processed_cur_selection)[1::2])\n",
```
gelzinyte, line 385: Do you think it's worth making (and how easy would it be) `ConfigSet` more list-like? In this case, make it suscriptable?

docs/source/examples.daisy_chain_mlip_fitting.ipynb
```diff
@@ -0,0 +1,726 @@
...
+    "isolated_atoms = generic.run(\n",
+    "    inputs=inputs,\n",
+    "    outputs=OutputSpec(),\n",
+    "    calculator=xtb_calc,\n",
+    "    properties=[\"energy\"],\n",
+    "    output_prefix=\"xtb_\"\n",
+    ")\n",
+    "\n",
+    "# set tag for fitting GAP\n",
+    "def modify_isolated_at(at):\n",
```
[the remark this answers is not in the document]
bernstei: I'm happy to add it back in.  It was originally intended for ABCD backend, but if people are using it, it's easy enough to add.

docs/source/examples.daisy_chain_mlip_fitting.ipynb
bernstei: I'll think about it.  It's coded as a generator because I really wanted to be able to use it for large datasets and not pull them all into memory at the same time, but that also prevents subscripting.

bernstei: Not as bad as I thought, if we're willing to accept that it could be inefficient, since it still has to go through all the configurations to the end (and this version, at least, doesn't support negative values like `-10:` for the last 10.

Can you want to try adding something this method to `ConfigSet`?
```
from itertools import islice
.
.
.
  def __getitem__(self, arg):
      if isinstance(arg, slice):
          return islice(self, arg.start, arg.stop, arg.step)
      elif isinstance(arg, int):
          return list(islice(self, arg, arg + 1, 1))[0]
      else:
          raise TypeError
```

docs/source/examples.daisy_chain_gap.ipynb
gelzinyte: I think the latter makes a bit more sense. 

While I think this would still be useful, I think this change is less crucial now that I've discovered `wfl.map`

docs/source/examples.daisy_chain_gap.ipynb
gelzinyte: I agree, there's one already in the docs explaining how input/output abstraction works.

docs/source/examples.daisy_chain_mlip_fitting.ipynb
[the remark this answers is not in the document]
gelzinyte: Thanks! Would you mind doing it? Here or in another PR?

docs/source/examples.daisy_chain_mlip_fitting.ipynb
gelzinyte: Had to modify the first return, but otherwise works!

bernstei: Is there a reason you converted the returned generators to lists? I don't generally like that, and the caller can always cast it into a list if they need an explicit list.

wfl/fit/error.py
gelzinyte: Ok, let me add this.

docs/source/examples.daisy_chain_gap.ipynb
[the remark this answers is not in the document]
gelzinyte: Ok, let me add that once tests without remote execution are fixed.

docs/source/examples.daisy_chain_gap.ipynb
bernstei: If c5a63de looks OK, can you resolve this conversation?

docs/source/examples.daisy_chain_gap.ipynb
bernstei: I also re-added `OutputSpec(... set_tags=dict)` in 5445057423f

docs/source/examples.daisy_chain_gap.ipynb
bernstei: I also re-added `OutputSpec(... set_tags=dict)` in 5445057423f

docs/source/examples.daisy_chain_mlip_fitting.ipynb
gelzinyte: It was throwing an error along the lines of "`islice` is not suscriptable". Withy my change it should be returning a `ConfigSet`. Please overwrite my changes if there's a better solution.

bernstei: If you want it to return *a ConfigSet* that's a different can of worms, which can also be done, but isn't what either of us implemented.  My version returns a single atomic config if the subscript is an int, or an iterable of configs is it's a slice. Yours returns a list if you pass a slice argument. Exactly what did you run that gave that error?  

If you tried to make it into a new `ConfigSet`, that wouldn't work, because `ConfigSet` needs a list, not an arbitrary iterable.  The current design is that you need to know that this constraint on creating `ConfigSet` exists, and that the subscript with a slice returns an iterable, not a list, and cast it yourself.  You can always iterate over the subscripts configset yourself and it'll work fine, or pass it to anything that can accept an iterable.

I can be convinced that the subscript with a slice should return a ConfigSet, or that you should be able to create a ConfigSet from an arbitrary iterable.  But returning a list strikes me as weird.  If we do decide that it should return a ConfigSet, we have to think about whether, if the original ConfigSet referred to files, whether the new one should refer to the same files but just skip some configs, or whether it should contain a distinct copy of the configs (presumably in memory).

docs/source/examples.daisy_chain_gap.ipynb
bernstei: OK - the immediate problem is that the header line is too long.  That's trivial to fix in `libAtoms`, but it doesn't actually help, because the fortran xyz reader only allows info float fields to be length 1, 3, 6, or 9. I think the new xyz reader fixes that, but I don't know how close we are to using it.  As a result, it chokes on the SOAP field either way.  I think we need to clean up the descriptor in this example, but I'll try to push QUIP to fix this issue.

bernstei:
> Again, this runs just fine executed as a notebook. So does the testing framework do some pickling which crash with workflow pickling? Or??

Have you run the pytest yourself, not in the CI? I'm trying, and been running for a couple of hours.  Is that reasonable?

bernstei:
Aha - the OpenMP hang issue.  Now I can reproduce the bugs the CI is hitting.

bernstei:
The undefined Atoms is is fixed by doing `exec(s, globals())`, as suggested in https://stackoverflow.com/questions/56683684/variables-in-list-comprehension-test-are-seen-as-not-defined-in-an-exec-with-pyt
I'll push an update.

docs/source/examples.daisy_chain_gap.ipynb
bernstei: OK, I fixed the random errors, but the test fails on the actual reference quantities. I suggest running it locally with pytest rather than relying on the CI, but I tried to give a more informative error. I'm not sure if the summary outputted by pytest is sufficient, but if you run pytest -s and scroll back you can find what actually failed. In my case it's a quantitative mismatch in the error table, specifically
```
E   AssertionError: Mismatch for ref prop atomization_energy/atom config train measure RMSE ref 0.014974721709894103 != actual 0.01606044820056456
```

bernstei:
OK, I fixed the random errors when pytest runs the docs examples, but the test (for the daisy chained workflow) fails on the actual reference quantities. I suggest you debug by running it locally with pytest rather than just relying on the CI to run it, but I tried to give a more informative error. I'm not sure if the summary outputted by pytest is sufficient, but if you run `pytest -s` and scroll back you can find what actually failed. In my case it's a quantitative mismatch in the error table, specifically
```
E   AssertionError: Mismatch for ref prop atomization_energy/atom config train measure RMSE ref 0.014974721709894103 != actual 0.01606044820056456
```
It appears to be non-deterministic for me, varying between 0.016 and 0.020 in a few tests.

docs/source/examples.daisy_chain_gap.ipynb
gelzinyte: Otherwise looks great, thanks!

docs/source/examples.daisy_chain_gap.ipynb
gelzinyte: Clearer reporting is very helpful, thanks. 

It seems that difference is because of different architectures gap_fit has been run on? The gap coefficients and predictions vary  _tiny_ bit from fit to fit, even with a given random seed. Also I get slightly different results when running notebook directly vs with pytest (which I didn't expect?). We could have CI values as reference, but then local tests would fail, so I'll relax the tolerance.

gelzinyte: It seems that the difference between local notebook and local pytest starts with RDKit geometries, even if they're generated the same whenever the notebook is re-run. I don't think I can set a random seed (Python RDKit is a wrapper to C++). Since this is mostly testing the workflow, I think very loose tolerance (0.1 eV or eV/Å) is the  optimal choice? This should pick out drastic breaking changes...

docs/source/examples.daisy_chain_gap.ipynb
bernstei: Sounds fine to me.  Can you do that change, and resolve this conversation?

docs/source/examples.daisy_chain_gap.ipynb
bernstei: Yeah, sometimes you need loose tolerances, or even testing some other (aggregate or averaged) quantity if you can't control some randomness.  We're having a bad time with that in ACEHAL, for reasons that I understand even less.

If you can reduce generating the initial non-deterministic configuration it to something simple, I can take a look as well.

gelzinyte: I have tried reading initial structures from file, but MD wasn't deterministic with set random seed. Or at least the notebook and pytest MD trajectories were different, didn't check if they were deterministic from run to run. I'm guessing that there's not enough significant figures to fill a double saved in xyz and the rest is filled with noise? 

Could try saving/loading the initial structures from e.g. ase.db, but not sure that's worth it?

docs/source/examples.daisy_chain_mlip_fitting.ipynb
gelzinyte: I think the (very slight) use-simplification I was looking for would mean making a ConfigSet from an arbitrary iterable, in addition to current `__getitem__`. But I don't think the added complexity is worth it after all (`list(iterable)` is perfectly fine). So let's leave it and I'll undo the changes from this thread?

bernstei: I agree with simpler.  

The ConfigSet from an arbitrary iterable isn't actually that complicated - the only problem is detecting what kind of input you passed `ConfigSet()`, but strings and `Atoms` are also iterable, so you can't just iterate and hope for the best. We could probably figure out sensible semantics if we wanted to.

docs/source/examples.daisy_chain_gap.ipynb
bernstei: The MD isn't deterministic if you're doing any parallelization unless you set the env var `WFL_DETERMINISTIC_HACK` to anything (which the gap rss iter fit test does, but isn't done across the board).  I guess we should consider fixing the non-determinism itself if possible, which I think I know how to do but it'd be tricky.

If that's the only issue, you could try setting that env var.

docs/source/examples.daisy_chain_mlip_fitting.ipynb
gelzinyte: I think let's separate this from this PR. I have opened issue #229. Please resolve this if that sounds good.

tests/test_doc_examples.py
```diff
@@ -14,22 +15,37 @@ def _get_coding_blocks(nb_file):
 
 
 @pytest.mark.parametrize(
-    ('nb_file', 'idx_execute'),
+    ('nb_file', 'idx_execute', 'needs_expyre'),
```
gelzinyte, line 18: @simonwengert what do you think about always executing all cells? In that case we can remove "idx_execute" option here?

simonwengert: The idea here was that some examples contain code-blocks that are not intended to be executed as python code (`cli` commands or `conda` installations, see https://libatoms.github.io/workflow/examples.smiles.html). If we want to integrate these examples in the auto-testing at some point, such code blocks can be skipped then easily. Although, we do not make use of this option, I think it's good to have it available.

gelzinyte:
Now everything is deterministic and error check passes with a default (tighter than before) threshold. Can this be merged @bernstei ?

bernstei:
I'm happy to merge, yes.

PR #232, opened 2023-03-11
Autopara per item info

bernstei:
Used to set per-item info, including number of config in overall sequence, as well as unique local rng seed, chosen well (hopefully) for parallel operation (see https://numpy.org/doc/stable/reference/random/parallel.html#sequence-of-integer-seeds).

Makes it possible to run deterministic  stochastic autopara ops (like md, for velocities, optimize, for random pressure, and minimahopping).

### DIFF
--- a/complete_pytest.tin
+++ b/complete_pytest.tin
@@ -4,7 +4,9 @@ module purge
 module load compiler/gnu python/system python_extras/quippy
 
 # should include both tin and tin_ssh - not as good as real remote machine, but maybe close enough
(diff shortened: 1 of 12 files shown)


gelzinyte:
Can this be aimed to be merged into `gap_fit_wfl_example`, not `main`?

bernstei:
> Can this be aimed to be merged into `gap_fit_wfl_example`, not `main`?

Fine by me.  Feel free to merge it into the wfl example branch, since tests are passing.

gelzinyte:
Thanks, I think this looks good, so I'll merge

PR #241, opened 2023-04-07
Some cli renames

gelzinyte:
Mainly create new command `calc` which houses sub-commands for calculating properties (energy/forces/.. with gap/aca/mace, atomization energy) and descriptors (quippy). I think this is slightly more intuitive. Any objections? Any other suggestions? @bernstei @simonwengert


wfl/cli/cli.py
```diff
@@ -33,31 +34,28 @@ def subcli_generate(ctx):
...
 subcli_select.add_command(cur)
 subcli_select.add_command(by_lambda)
 
-
-@cli.group("eval")
+@cli.group("calc")
 @click.pass_context
-def subcli_eval(ctx):
+def subcli_calc(ctx):
+    "Calculate properties and descriptors."
```
bernstei, line 47: Not that it matters, but `"""` to keep consistency with the other docstrings?

wfl/cli/commands/calc.py
```diff
@@ -93,7 +94,8 @@ def mace(ctx, inputs, outputs, param_fname, prop_prefix, num_inputs_per_python_s
     help="``atoms.info`` key on which to select isolated atoms")
 @click.option("--isolated-atom-info-value", "-v", default="default", 
     help="``atoms.info['isolated_atom_info_key']`` value for isolated atoms. Defaults to \"IsolatedAtom\" or \"isolated_atom\"")
-def atomization_energy(inputs, outputs, prop_prefix, prop, isolated_atom_info_key, isolated_atom_info_value):
+def atomization_energy(ctx, inputs, outputs, prop_prefix, prop, isolated_atom_info_key, isolated_atom_info_value):
+    """Calculte atomization energy"""
```
bernstei, line 98: typo

wfl/cli/commands/calc.py
```diff
@@ -104,3 +106,22 @@ def atomization_energy(inputs, outputs, prop_prefix, prop, isolated_atom_info_ke
     )
 
 
+@click.command("quippy")
```
bernstei, line 109: Is this already under "descriptors" somehow?  If not, we should possibly move it under such a category, first because quippy also provides potentials, and 2nd, because people might want to calculate ACE descriptors some day.

gelzinyte: Ah yes, this is the reason for this renaming. Currently I've merged what was in "eval" and "descriptors" into "calc", because things in both categories could be "calculated" (@gabor1 pointed out). Didn't think about quippy/ace doing both, properties and descriptors. Maybe instead should rename "eval" -> "calc-prop" & "descriptors" -> "calc-desc" or "calc-properties" & "calc-descriptors"? Just `wfl descriptors` would be fine, but `wfl properties` is maybe not as clear? Or should we just keep `eval` and `descriptors`? (It's not a big issue, but I thought I'd touch on this in the same PR"

bernstei: So this is now "calc quippy"?  Can you briefly list the relevant part of the nested command tree structure that's available now (as of this PR)?

gelzinyte: yes; edited the PR description

bernstei:
what about moving quippy from "calc" to a new top level category "descriptors"?

gelzinyte:
I guess the only shortcoming is that descriptors are also "calculated", but if that's fine with you I'll make the change.

bernstei:
On slack @gabor1 suggests longer names, so how about `calculators` and `descriptors`?

gelzinyte:
Done!

gelzinyte:
Is this ok to merge? The command tree in the initial PR message is what this PR implements.

bernstei:
@gelzinyte trying to do some cleanup.  Do we remember what's the status of this one, esp. with respect to what we described in the paper?

PR #247, opened 2023-06-08
Docs caching warning

bernstei:
Add warnings in the docs about ways the caching of information (output files from all autoparallelized operations and stage directories from expyre jobs) can fail if the source code that's being executed is changed.


bernstei:
@gelzinyte can we merge this one too, since we're referring to it in the revised paper?

gelzinyte:
This can be merged too, thanks!

PR #256, opened 2023-08-08
Debug of per config calculator

Felixrccs:
I tried the branch out and run into problems:

1.  The `calculator` was either a `pickled_calculator` or `None`, so `calculator[0]` does not work. (I added a variable calculator_info, which is probably not the most elegant Solution.)
2.  Missing brackets for the construct_calculator_picklesafe.

This version is now working and tested on my setup.

### DIFF
--- a/wfl/calculators/generic.py
+++ b/wfl/calculators/generic.py
@@ -46,6 +46,7 @@ def _run_autopara_wrappable(atoms, calculator, properties=None, output_prefix='_
     if properties is None:
         properties = ['energy', 'forces', 'stress']
     try:
+        calculator_info = calculator
         calculator = construct_calculator_picklesafe(calculator)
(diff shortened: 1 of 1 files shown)


bernstei:
>  (I added a variable calculator_info, which is probably not the most elegant Solution.)

What do you think about getting rid of your new `calculator_info`, but instead saving the returned value of `construct_calculator_picklesafe` in `calculator_default` (not overwriting `calculator`), leaving `calculator` with the original value that was passed into the generic calculate function?

Felixrccs:
That sound good to me, I can try that out.

Felixrccs:
I will trie the changes tomorrow morning.

Felixrccs:
Ok this should do the trick, I tried it with VASP and remote again and it works fine.

PR #261, opened 2023-08-21
Per configuration pytest

Felixrccs:
First trial, lets see if this works.

### DIFF
--- a/tests/prep_test_cli_rss.sh
+++ b/tests/prep_test_cli_rss.sh
@@ -19,7 +19,7 @@ fi
 
 export VASP_COMMAND=vasp.serial
 export VASP_COMMAND_GAMMA=vasp.gamma_serial
-export PYTEST_VASP_POTCAR_DIR=${VASP_PATH}/pot/rev_54/PBE
+export VASP_PP_PATH=${VASP_PATH}/pot/rev_54/PBE
 export GRIF_BUILDCELL_CMD=${HOME}/src/work/AIRSS/airss-0.9.1/src/buildcell/src/buildcell
 
 nohup pytest -s tests/test_cli_rss.py 1> prep_test_cli_rss.stdout 2> prep_test_cli_rss.stderr &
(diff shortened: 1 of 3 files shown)


Felixrccs:
Ok I think something is not working with the vasp tests. (all of them are skipped due to a missing environment variable)

Felixrccs:
I found an error in the `tests/prep_test_cli_rss.sh`. But in the end the VASP path's are not declared and the pytest is always skipped. I would not be surprised if most of the VASP tests fail if we get this to work again.

Im not familiar how these test pipelines are set up. So I need some help there.

bernstei:
> I found an error in the `tests/prep_test_cli_rss.sh`. But in the end the VASP path's are not declared and the pytest is always skipped. I would not be surprised if most of the VASP tests fail if we get this to work again.

There's no VASP on the github CI, since it requires a license.  Someone has to run the tests manually. I have a local script that does that by setting various env vars. Generally I've been the only one really using VASP a lot, so I've been running that test when anyone submits a PR that's Vasp related.  I'm happy to do that for this PR, but it'd also be nice to have a test that's _not_ Vasp-based, so it'll run on the github CI as well.

Felixrccs:
Ok I'll try to do a non VASP pytest.

bernstei:
Feel free to keep the Vasp test in there as well, and I'll run it locally (in addition to the CI running the non-Vasp one).

tests/prep_test_cli_rss.sh
```diff
@@ -19,7 +19,7 @@ fi
 
 export VASP_COMMAND=vasp.serial
 export VASP_COMMAND_GAMMA=vasp.gamma_serial
-export VASP_PP_PATH=${VASP_PATH}/pot/rev_54/PBE
```
bernstei, line 22: Why did you change this variable? It's internal to ASE's vasp calculator, and this isn't even a test - it's a script that does the reference gap_rss_iter_fit calculation which the corresponding test compares to.

bernstei: Do we need a more general discussion of how the VASP tests are configured?

Felixrccs: I first thought that was part of the reason why the pytest for VASP was skipped.
So this change make no sense after since I now that the pytest is not done in the pipeline.

Issue #253, opened 2023-07-21
autoparallelization over different calculation setting


jungsdao:
Hello, 
I'd like to ask how I can parallelize over different computational settings with the same geometry. 
I think it's quite understandable how to perform high-throughput calculation with fixed calculation setting. 
However, it's less intuitive with the other way around. (For example looking for k-points convergence or different energy cutoff values with the same structure.)
If such screening is possible within current setting, could you guide me with a snippet of code that carries out such task? 
Many thanks in advance!

Best regards, 
Hyunwook

bernstei:
That's not really anything we've currently designed to support, although we've talked about it. We can look into how hard it'd be.

bernstei:
I looked over the code, and I can think of a a fairly straightforward way to do it.  

We could pass a per-atomic-configuration `args` and/or `kwargs` in the `Atoms.info` field, and if we detect those use them to override the calculator that's passed in.  I guess that'd require passing in the calculator constructor as a triple (rather than an already created calculator object), but that seems manageable to me.  It would autoparallelize naturally, since really the parallelization is over the atoms objects.

bernstei:
@gabor1 @gelzinyte do we think this use case is worth it?  If what I write above is all that's needed, it's not much work.  Just a couple of if statements and moving the calculator constructor inside the atoms loop.

 If it leads to some subtle issue, e.g. multiple autoparallelized calculations conflicting with each other, it could be tricky, but I don't think that'll happen.

gabor1:
I think it's a nice use case, and as you say generalises the pairing of calculators and atoms objects.

gelzinyte:
I've heard from other people too that this feature would be useful, so I think it's very much worth implementing it.

bernstei:
OK - I'll put together a PR, although i don't know exactly how long it'll take.  Probably a couple of weeks until it's ready to reveal publicly.

bernstei:
#254 should do this.  Would someone be willing to test it, ideally by making a pytest that uses the functionality?

bernstei:
closed by #254

PR #262, opened 2023-08-28
Add universal kspacing and refactor handling of nonperiodic cells

### LINKED ISSUE #258: The Vasp calculator with pbc = False runs VASP multiple times
The wfl Vasp calculator wrapper changes all the pbcs to `True` before actually running VASP since the ASE Vasp calculator complains otherwise, and then changes back once the calculation is done.  However, the saving of the properties triggers additional VASP runs, basically because the ASE system detects the change in pbc values as a change to the system and invalidates the cache when `save_results()` calls things like `get_potential_energy()`. 

For VASP it'd be easy enough to make `save_results()` try to grab the values from `atoms.calc.results`, which won't trigger another evaluation. I'm not sure how that'd behave with CASTEP, which doesn't have a `Calculator.results` dict.

### LINKED ISSUE #259: DFT calculators do all sorts of different things with nonperiodic systems
In the process of implementing a universal kspacing definition, I discovered that the various DFT calculators do all sorts of different things when they encounter non-periodic systems.  It would be nice if all these behaviors were rationalized and made consistent, although it's possible that the input params are so different that this won't actually achieve much. But I think it's worth a review.

bernstei:
Universal kspacing only implemented in Vasp so far.

Refactoring of nonperiodic mainly consists of calculators.generic doing the removal of stress for fully non-periodic, and cleaning up the DFT calculators to only do the minimum.

Also, the saving of the results can optionally be done in the calculator, and Vasp uses that because it changes the pbc, and that can break the ASE result caching and lead to extra calculations (#258).

closes #258 
closes #259


bernstei:
@Felixrccs this one is built on top of the per-configuration calculator params, so we can merge it once #253 (with its pytest) is merged. Also, if we want the universal k-point spacing definition to apply to calculators other than VASP, it'd have to be implemented in each.

Felixrccs:
Sounds great I hope I‘ll get the pytest done at the beginning of next week.

Felixrccs:
> What do you mean by "there"? I already put it in Vasp in this PR. Did you mean add it to some other calculator?

sorry that was a typo I meant the QE calculator

bernstei:
Just merged from main which had the per-config calculator stuff merged.  Assuming it still passes, how much more work do we want to do before we merge?  Add QE (@Felixrccs ?) or CASTEP support? Add a test?

Felixrccs:
I try to work on QE, but next weeks are busy so it will take some time.
If we wait until we manage to add everything I can also look in test cases again.

wfl/calculators/espresso.py
```diff
@@ -123,54 +126,46 @@ def calculate(self, atoms=None, properties=_default_properties, system_changes=a
...
+            self.parameters["kspacing"] = None
+            self.parameters["koffset"] = False
+        elif np.any(~self.atoms.pbc):
+            # mixed T & F
+            if "kspacing" in self.parameters:
+                # need to create the grid,
+                # `kspacing` overwrites `kpts`
+                # we set it in there and
+                self.parameters["kpts"] = kspacing_to_grid(
+                    self.atoms, spacing=self.parameters["kspacing"] / (2.0 * np.pi)
```
Felixrccs, line 147: @bernstei Do we want to stick with the `ase.io.espresso kspacing_to_grid` function in the qe calculator or do we switch to the `universal_kspacing_n_k`? I think switching it could be good to make the code more uniform. However as far as I've seen the codes both function have basically the same function.

bernstei: I think we should use the same function for all the calculators, so it's actually universal, but I'm ambivalent about doing our own vs. using ASE's (the usual tradeoff of control vs. not doing extra work). If they give the same results, we can use the ASE for now, and switch back to our own implementation if we need something more, or if they break it.

bernstei:
@Felixrccs where are we at with this one?

Felixrccs:
I still have it on my mind but I did not menage to work on it yet.

bernstei:
@Felixrccs still thinking about this one?

Felixrccs:
I still have it on my mind but I cannot say when I get enough free time to continue work on this. I'll try to further work on this but I dont think Ill manage it before Christmas.

bernstei:
OK.  Once the tests pass (failing due to unrelated ASE updates), I'll merge.  Off topic, I wonder if we should switch from testing against the latest github ASE to a specific release, since ASE seems to be changing their API a fair amount lately.

PR #264, opened 2023-08-31
Global opt using ASE Minima Hopping

jungsdao:
(Perhaps) low-level modifications to wfl package.


bernstei:
The diffs show that this PR is reverting changes we've made to main (the one I noticed was the reversion of `calculate()` to `run()` in `calculators.generic`, from 4 months ago).  I'm not sure how that happened give that you seem to have merged main into your branch very recently, but all those need to be fixed before we can really consider the merge.

wfl/fit/error.py
```diff
@@ -216,7 +216,7 @@ def _reshape_normalize(quant, prop, atoms, per_atom):
         for cat in all_diffs[prop]:
             diffs = np.asarray(all_diffs[prop][cat])
             weights = np.asarray(all_weights[prop][cat])
-
+
```
bernstei, line 219: Please clean up changes that are just whitespace.

wfl/fit/gap/multistage.py
```diff
@@ -82,6 +83,28 @@ def _select_info(ats, info_keys):
                 del at.info[k]
 
 
+def get_Num_bonds(slab):
```
bernstei: Why is this function in `gap/multistage.py`? Also, it'd almost certainly be much faster to use the linear scaling ASE neighborlist https://wiki.fysik.dtu.dk/ase/ase/neighborlist.html#ase.neighborlist.neighbor_list, or even better the matscipy one which has identical semantics (although different spelling of the functions)

wfl/select/stratify.py
```diff
@@ -0,0 +1,91 @@
+import warnings
+
+import numpy as np
+import random
+
+
+def test_short_distance(structure, mindist = 0.1):
```
bernstei, line 7: Why is this in a file called `stratify.py`?

wfl/fit/error.py
jungsdao: I don't understand why this is counted as 'diff' here. I don't see any difference between this script and one from the main. It has the same white space.

wfl/fit/gap/multistage.py
jungsdao: This part has been removed from multistage.py file

wfl/fit/gap/multistage.py
```diff
@@ -97,7 +120,7 @@ def max_cutoff(params):
 
 # noinspection PyPep8,PyPep8Naming
 def fit(fitting_configs, GAP_name, params, ref_property_prefix='REF_',
-        seeds=None, skip_if_present=False, run_dir='.',
+        seeds=None, skip_if_present=False, run_dir='.', use_heuristics = False,
```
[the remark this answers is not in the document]
jungsdao: This argument has been also removed.

wfl/fit/gap/multistage.py
```diff
@@ -389,10 +414,11 @@ def fit(fitting_configs, GAP_name, params, ref_property_prefix='REF_',
                         print('stage {} config {} residual error {} {}'.format(i_stage, at.info.get('config_type'),
                                                                                Egap / len(at), Eref / len(at)))
 
-        abs_dE_sum = np.sum(np.abs(dEs))
-        print('energy error MAE per atom {} per descriptor {}'.format(abs_dE_sum / np.sum(at_Ns),
-                                                                      abs_dE_sum / descriptor_count))
-        delta = abs_dE_sum / descriptor_count
+		abs_dE_sum = np.sum(np.abs(dEs))
```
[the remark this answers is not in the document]
jungsdao: I don't understand why this is denoted as diff either

wfl/fit/error.py
bernstei: tab vs. spaces?

wfl/fit/mace.py
```diff
@@ -0,0 +1,124 @@
...
+from expyre import ExPyRe
+from wfl.configset import ConfigSet, OutputSpec
+from wfl.autoparallelize.utils import get_remote_info
+from expyre.resources import Resources
+from pathlib import Path 
+from shutil import copyfile
+
+
+
+def run_mace_fit(params, mace_name="mace", run_dir=".", remote_info=None, mace_fit_cmd="python  ~/Softwares/mace/scripts/run_train.py",
```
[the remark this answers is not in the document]
jungsdao: I wasn't planning to include mace.py into this pull request because I'm still working on this, but accidentally added to here. But I'll change this.

wfl/select/stratify.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    distancematrix = structure.get_all_distances()
+    distancematrix[np.diag_indices(distancematrix.shape[0])] = 1
+    if np.min(distancematrix) < mindist:
+        return False
+    else:
+        return True
+
+
+def stratified_random(inputs, outputs, num=5):
```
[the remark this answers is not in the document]
bernstei: That's a much more specific thing than just "stratifying" implies.  For this to be more generally useful I think it would need to be generalized.

One possibility is to write a general selection method that picks some number of configs from each category (the usual meaning of stratify is to subselect keeping the same relative population in each category, which isn't what you want here), with the category coming from an arbitrary `Atoms.info` (could default to `"config_type"`), and something like a dict with the keys being the category values and the values being the number of configs to select (or a fraction, with the total # in a separate argument?).

Then your top level script could make sure that the info field contained the category as you define it (global vs. local vs. random MD), and passes the dict and/or total number to get the 5 configs you want.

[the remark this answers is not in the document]
bernstei: Note that I haven't looked at all as to whether any of the existing selection methods do any of this sort of thing.

wfl/generate/md/__init__.py
```diff
@@ -136,19 +137,30 @@ def _sample_autopara_wrappable(atoms, calculator, steps, dt, temperature=None, t
...
+        if temperature_tau is None and friction is None:
             # NVE
             if pressure is not None:
                 raise RuntimeError('Cannot do NPH dynamics')
             md_constructor = VelocityVerlet
             # one stage, simple
             all_stage_kwargs = [stage_kwargs.copy()]
             all_run_kwargs = [ {'steps': steps} ]
+
+        elif temperature_tau is None and friction is not None:
```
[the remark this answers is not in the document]
jungsdao: I don't know how I can relate `temperature_tau` and `friction`.
It was easy way to include Langevin dynamics into the MD script. 
I think larger modification might be needed in order to make it to specify which thermostat to be used.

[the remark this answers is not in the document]
bernstei: Well, I definitely don't think we should have two different arguments that aren't can't be specified at the same time and have it behave unexpectedly if you pass both. tau = 1 / friction.

wfl/select/stratify.py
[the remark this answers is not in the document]
jungsdao: I heard that `by_bool_func` in `select/simple.py` can actually accommodate this kind of sampling. I wasn't aware of it. So I'll just remove `stratify.py` from this PR and this sampling could be rather added to top-level code which is more specific to the individual project.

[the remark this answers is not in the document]
bernstei: I don't really see how `by_bool_func` can do this sampling, because I thought it just makes a yes/no decision about each config independent of all the others, but I'm happy to leave it up to you how you do the sampling you need for your top level code. I just want anything that gets incorporated into wfl itself to be at least somewhat general.

wfl/generate/md/__init__.py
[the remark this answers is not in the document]
jungsdao: Following some advice from Elena, `__init__.py` in md has been restructured by adding new argument `integrator`. This argument has to be either `default` or `langevin` and if one chooses langevin, it will get friction value from `1/temperature_tau`. However, I think it could be more intuitive for users to use `temperature_tau` rather than `1/temperature_tau`.

wfl/fit/error.py
jungsdao: I hope this problem is resolved in current script.

wfl/fit/gap/multistage.py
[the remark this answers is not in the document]
jungsdao: I hope this problem is resolved in current script as well

jungsdao:
Hello, is there any further things that should be done from my side?
I just wanted to ask whether this pull request can be proceeded. Thank you :)

wfl/generate/md/__init__.py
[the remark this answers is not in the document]
bernstei: Why `"default"`, rather than `"berendsen"` (or maybe the full ASE name `"NVTBerendsen"`), and just make that the default value?

wfl/fit/gap/multistage.py
```diff
@@ -394,6 +394,7 @@ def fit(fitting_configs, GAP_name, params, ref_property_prefix='REF_',
                                                                       abs_dE_sum / descriptor_count))
         delta = abs_dE_sum / descriptor_count
 
+
```
bernstei, line 397: remove extra blank line

wfl/fit/mace.py
```diff
@@ -0,0 +1,153 @@
+import numpy as np
```
bernstei, line 1: I'd like to have a test for this functionality.  I guess it's a separate issue as to whether it can be run automatically by the github CI, but at least one that people can run manually.

wfl/generate/md/__init__.py
```diff
@@ -43,7 +46,7 @@ def _sample_autopara_wrappable(atoms, calculator, steps, dt, temperature=None, t
         - [ {'T_i': float, 'T_f' : float, 'traj_frac' : flot, 'n_stages': int=10}, ... ] list of stages, each one a ramp, with
         duration defined as fraction of total number of steps
     temperature_tau: float, default None
-        time scale that enables Berendsen constant T temperature rescaling (fs)
+        time scale that enables Berendsen constant T temperature rescaling (fs). If integrator is set to Langevin, friction is 1/temperature_tau.
```
bernstei, line 49: `Time scale for thermostat.  Directly used for Berendsen integrator, or as 1/friction for Langevin integrator`

wfl/generate/minimahopping.py
```diff
@@ -2,100 +2,208 @@
...
 from wfl.autoparallelize import autoparallelize, autoparallelize_docstring
 from wfl.utils.misc import atoms_to_list
 from wfl.generate.utils import config_type_append
 from wfl.utils.parallel import construct_calculator_picklesafe
 
 
+def _get_MD_trajectory(rundir):
+	
+	md_traj = []
+	mdtrajfiles = sorted([file for file in glob.glob(f"{rundir}/md*.traj")])
```
bernstei, line 21: I have a slight preference for using `pathlib` when possible, so here that'd be `Path(rundir).glob("md*.traj")`.

wfl/generate/md/__init__.py
```diff
@@ -34,6 +35,8 @@ def _sample_autopara_wrappable(atoms, calculator, steps, dt, temperature=None, t
         ASE calculator or routine to call to create calculator
     dt: float
         time step (fs)
+	integrator: str
```
bernstei, line 38: The docstring syntax for defaults is
```
integrator: str, default "Berendsen"
```

wfl/generate/md/__init__.py
```diff
@@ -78,6 +81,8 @@ def _sample_autopara_wrappable(atoms, calculator, steps, dt, temperature=None, t
     -------
         list(Atoms) trajectories
     """
+	
+    assert integrator.lower() in ["default", "langevin"]
```
bernstei, line 85: Please drop the `.lower()` calls and just make integrator be case sensitive?  Everything else in python is.

Also, I think it's probably worth using the full ASE class names, i.e. `Langevin` and `NVTBerendsen`?

bernstei:
> Hello, is there any further things that should be done from my side? I just wanted to ask whether this pull request can be proceeded. Thank you :)

I closed the old comments, and did another pass.  Mostly small stylistic things, except that I'd really like to have a test for the mace fit, on some tiny model, just to make sure the syntax is correct.

wfl/generate/minimahopping.py
```diff
@@ -2,100 +2,208 @@
...
 
+#        else:	
+#            print("Parallel run returns minima.traj", flush=True)
+#            os.chdir(workdir)
+#            return None
 
-def _run_autopara_wrappable(atoms, calculator, Ediff0=1, T0=1000, minima_threshold=0.5, mdmin=2,
-                           fmax=1, timestep=1, totalsteps=10, skip_failures=True,
+
+def _run_autopara_wrappable(atoms, calculator, Ediff0=1, T0=1000, minima_threshold=0.5, mdmin=2, parallel=1,
```
bernstei, line 152: Why is there not docstring for `parallel`?

bernstei:
In addition to the little comments, there's a major issue which is that the assumption behind the `wfl` autoparallelization is that each operation is entirely separate.  This means that you can't really have something like `parallel` - since `MinimaHopping` writes files, just run each one in a separate temporary directory the way the DFT calculators do.  

Also, there are various commented out things that seem to just be old code, and lots of things are missing from docstrings (arguments from ones that are there, and no docstrings at all for a number of the functions). Can you clean all those up before we try to do a complete review of the `minimahopping` wrapper?

bernstei:
I also strongly suggest not using `autopara_per_item_info` except for setting the random seed the way `md.py` uses it.  It's a, kind of a horribly messy hack, and b, not guaranteed to be supported in any particular way in the future.

wfl/generate/minimahopping.py
```diff
@@ -2,100 +2,208 @@
...
+        if parallel == 1:
+
+            if return_all_traj:
+                traj += _get_MD_trajectory(rundir)
+            for hop_traj in Trajectory('minima.traj'):
+                config_type_append(hop_traj, 'minima')
+                traj.append(hop_traj)
+
+        elif parallel > 1 and parallel_seed is not None:
+            print(f"parallel run {parallel_seed} suceeded.")
```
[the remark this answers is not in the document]
jungsdao: It's a bit tricky in parallel minima hopping. Although each minima hopping is carried out individually, the outcome of whole subprocesses is `minima.traj` which is shared by all different subprocesses. So individual run returns None, but the user can get one `minima.traj` .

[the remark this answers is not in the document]
bernstei: That's not the way `wfl` works.  If the user wants to take the union of the `minima.traj` that are associated with each of the input configs, that's fine.  But the `wfl` autoparallelization assumption is that the outputs should be mapped one-to-one with the inputs.

The way ASE implements it, writing to a file with a fixed name, is very limiting, especially if you want to parallelize.  I don't see any usefulness in `wfl` replicating that, when it's so easy to do it sensibly, and then the user can take the union of all the outputs if they want. It should run minimahopping for each input config in its own directory, read the files that it returns, and return whatever you think is useful (minima, optimzation trajectories, etc) as the op output.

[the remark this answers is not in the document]
bernstei: Note that I'm assuming that the combined `minima.traj` is only _written_ to by the various parallel tasks to combine the outputs. If the updated file is also read back in for some reason during the minima hopping process, that's another can of worms, because it's no longer embarrassingly parallel. Relying on things like shared files output will break, e.g., if the inputs are split into multiple remote jobs that may not run on the same machine, or at the same time.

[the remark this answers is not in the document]
jungsdao: Okay, after understanding that, I think maybe parallel minima hopping might not be suitable to `wfl` frame which could be more suitable to high-level features. If this is too complicated to be incorporated into main branch, I'll rather just drop updates in `minimahopping.py`. What do you think?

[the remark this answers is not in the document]
bernstei: I don't have any strong feelings.  If you think running the separate initial configs entirely separate is not a useful behavior, then we should probably drop it from `wfl`.  However, I want to point out that what you're doing when you just submit entirely separate runs that share a single `minima.traj` file is not reliable.  I looked over the code, and I don't see any evidence that it's written to handle that situation.  What if two processes try to write the new config at the same time? 

If you want multiple searches sharing a single set of minima safely you'd have to write that with explicit MPI parallelization, and while not that complicated, it wouldn't be trivial, because each search can take a different amount of time. One thing you might be able to do is have each search write to its own file, but when they read back the minima they'd read all the files, from all the tasks. That might still have problems if one task tries to read _while_ another is writing, but you can even get around that by writing to a temporary file and renaming it to the normal name once the write is complete.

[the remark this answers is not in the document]
jungsdao: Those potential errors involved in sharing a file between different processes are something I haven't expected. I think implementing MPI parallelized minima hopping is beyond my work scope and there's already another implementation done by Stefan Goedecker's group. Trying this would be quite redundant. So I would rather drop parallel minima from `wfl` and make it rather top-level thing.

wfl/fit/mace.py
jungsdao: If I want to make pytest for this, pytest file should know where the `run_train.py` file is located. And this can be accessed by tracing back where the dist-info directory is. From there maybe I can look into `direct_url.json` file where it records the location of mace directory with `run_train.py` is located. So it can be done like this but I'm not totally sure whether this is proper way. Otherwise one has to specify where the path of `run_train.py` .

bernstei: You can use 
```
import mace
mace_dir = Path(mace.__file__).parent
```
for example, if you know that `run_train.py` is installed in the same directory as the mace python package (although it would be nicer if mace installed `run_train.py` as a proper binary, so it was just in your path, but obviously that's not something this PR can fix. I can try to convince them to do that).

If you write the pytest, I can help polish it so it runs cleanly on different machines.  Let me look into how the mace pip installation works.

wfl/generate/minimahopping.py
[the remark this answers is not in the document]
bernstei: That's fine by me.  I'll resolve this conversation, and we can keep the remainder (Langevin integrator, mace fitting)

wfl/fit/mace.py
jungsdao: I've pushed a simple pytest for mace fitting. Please take a look and let me know if there's some room for improvement.

bernstei: Thanks.  I'll take a look, and see if I can improve the mace pip installer if that's needed as well.

wfl/generate/md/__init__.py
```diff
@@ -149,20 +154,28 @@ def _sample_autopara_wrappable(atoms, calculator, steps, dt, temperature=None, t
...
             else:
-                md_constructor = NVTBerendsen
+                if integrator == "NVTBerendsen":
+                    md_constructor = NVTBerendsen
+                    stage_kwargs['taut'] = temperature_tau * fs
+
+                elif integrator == "Langevin":
+
+                    md_constructor = Langevin
+                    stage_kwargs["friction"] = 1 / temperature_tau
```
bernstei, line 178: Any reason why you aren't doing the fs unit conversion here as well (although since ASE's native units are fs, it doesn't actually do anything)?

wfl/generate/minimahopping.py
```diff
@@ -1,107 +1,107 @@
...
-
-
-# run that operation on ConfigSet, for multiprocessing
-def minimahopping(*args, **kwargs):
-    default_autopara_info = {"num_inputs_per_python_subprocess": 10}
-
-    return autoparallelize(_run_autopara_wrappable, *args,
-                           default_autopara_info=default_autopara_info, **kwargs)
-autoparallelize_docstring(minimahopping, _run_autopara_wrappable, "Atoms")
+import os
```
bernstei, line 1: What's the final decision about minimahopping? Is this just back to some old version,  or did you decide to implement it without sharing a single minima file between the different autoparallelized tasks?

bernstei:
Can you switch all your use of `os.path` to `Path` (from `pathlib`), (also use `Path`'s `glob` method rather than `glob.glob`, although not if it's not straightforward) and also use the overloaded `/` operator, i.e.
```
   run_dir / "subdir" / "filename"
```
rather than
```
   run_dir / "subdir/filename"
```

wfl/generate/minimahopping.py
jungsdao: I'll rather just drop parallel minima hopping part but I'd like to update one  thing if possible. In old wfl implementation, minima hopping is ran in tmp folder and erased right after running. Also it returns only minima structures.

But I think it would be useful to keep the trajectories and return MD structures as well for iterative MLP training. So if you think this would be useful as well, I'll update only this feature and push again.

bernstei:
Can you make the mace fitting code be more like, e.g. the ace fitting code?  Same argument names when possible, same order of arguments, make a `fit` function (you can keep the executable functionality if you want, but I personally don't think it's useful), etc.  Also, make the indentation 4 spaces rather than 8.

wfl/generate/minimahopping.py
bernstei: That'd be fine by me. You can also take a look at how the DFT calculators do the tmp run directory cleanup, if you want to see how we do it elsewhere in wfl.

jungsdao:
> Can you make the mace fitting code be more like, e.g. the ace fitting code? Same argument names when possible, same order of arguments, make a `fit` function (you can keep the executable functionality if you want, but I personally don't think it's useful), etc. Also, make the indentation 4 spaces rather than 8.

What do you mean indentation 4 spaces rather than 8? Could you give me some example?

bernstei:
> What do you mean indentation 4 spaces rather than 8? Could you give me some example?

I just realized that it's not 8 spaces, you're using a tab for indentation. That's not recommended in modern python (https://peps.python.org/pep-0008/#tabs-or-spaces), and we use 4 spaces for each level of indentation in all of the rest of the code.

bernstei:
> What should `fit` function do? From `ace.py`, It seems just preprocessing before actual fitting. Also currently mace_fitting does not accept list of ASE objects but rather accepts path of `xyz` file. I'm not sure it's okay to go as it is.

I don't have that strong of an opinion on how it should be split into functions inside, but I just want the function names and functionalities consistent with other fitting code wrappers in wfl. So I want the function that's called by the user to be named `fit`, and for it to accept a `ConfigSet` and write it to a file, same as the other fitting wrappers. Whenever there's a function argument that's equivalent it'd be good to be named the same, unless there's a specific reason not to.

wfl/generate/minimahopping.py
jungsdao: I have added arguments to `minimahopping.py` that includes md trajectory along with local minima in the output.
Also there are some corresponding modifications in `test_minimahopping.py` as well.

jungsdao:
> Can you make the mace fitting code be more like, e.g. the ace fitting code? Same argument names when possible, same order of arguments, make a `fit` function (you can keep the executable functionality if you want, but I personally don't think it's useful), etc. Also, make the indentation 4 spaces rather than 8.

I have replaced all the tab to 4 spaces. I hope there's any file left with tab indentation.

jungsdao:
> > What should `fit` function do? From `ace.py`, It seems just preprocessing before actual fitting. Also currently mace_fitting does not accept list of ASE objects but rather accepts path of `xyz` file. I'm not sure it's okay to go as it is.
> 
> I don't have that strong of an opinion on how it should be split into functions inside, but I just want the function names and functionalities consistent with other fitting code wrappers in wfl. So I want the function that's called by the user to be named `fit`, and for it to accept a `ConfigSet` and write it to a file, same as the other fitting wrappers. Whenever there's a function argument that's equivalent it'd be good to be named the same, unless there's a specific reason not to.

I also have rearranged arguments in  `mace.py` and modified pytest. I wonder if current status is good enough. Feel free to give me feedback :)

bernstei:
> I also have rearranged arguments in `mace.py` and modified pytest. I wonder if current status is good enough. Feel free to give me feedback :)

I'll take a look, hopefully later today, otherwise tomorrow.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+        remote_info = get_remote_info(remote_info, remote_label)
+    
+    
+    if remote_info is not None and remote_info != '_IGNORE':
+        input_files = remote_info.input_files.copy()
+        output_files = remote_info.output_files.copy() + [str(run_dir)]
+    
+    
+        # set number of threads in queued job, only if user hasn't set them
+        if not any([var.split('=')[0] == 'WFL_GAP_FIT_OMP_NUM_THREADS' for var in remote_info.env_vars]):
```
bernstei, line 90: If MACE fitting actually uses OpenMP, these should be changed to `WFL_MACE_FIT_OMP_NUM_THREADS`, or if not, all this stuff should be removed.

wfl/generate/minimahopping.py
```diff
@@ -1,107 +1,134 @@
...
+from wfl.utils.parallel import construct_calculator_picklesafe
+
+
+
+def _get_MD_trajectory(rundir):
+    
+    md_traj = []
+    mdtrajfiles = sorted([file for file in Path(rundir).glob("md*.traj")])
+    for mdtraj in mdtrajfiles:
+        for at in ase.io.read(f"{mdtraj}@:"):
```
bernstei, line 24: Can you switch from `read("file.xyz@:")`, which is deprecated in ASE, to `read("file.xyz", ":")` ? There's also a place in the mace fitting test, and possibly elsewhere.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+        directory to run fitting in
+    remote_info: dict or wfl.autoparallelize.utils.RemoteInfo, or '_IGNORE' or None
+        If present and not None and not '_IGNORE', RemoteInfo or dict with kwargs for RemoteInfo
+        constructor which triggers running job in separately queued job on remote machine.  If None,
+        will try to use env var WFL_EXPYRE_INFO used (see below). '_IGNORE' is for
+        internal use, to ensure that remotely running job does not itself attempt to spawn another
+        remotely running job.
+    verbose: bool default True
+        verbose output
+    do_fit: bool, default True
```
bernstei, line 47: already using `dry_run` (with the opposite sense) for this behavior.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+        carry out the fit, otherwise only print fitting command 
+    wait_for_results: bool, default True
+        wait for results of remotely executed job, otherwise return after starting job
+    remote_label: str, default None
+        label to match in WFL_EXPYRE_INFO
+    skip_if_present: bool, default False
```
bernstei, line 53: please make docstring arguments consistent (the same order, nothing missing, nothing extra) as the arguments in the function declaration. Specifically, I noticed that `kwargs` is undocumented (maybe other things, too - it's hard to tell when they're out of order.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+        configurations to fit to
+    use_params: dict
+        ACE fit parameters, will have input filename set based on where configs were written to
+    ace_file_base: str
+        base to all ACE-related files, used for saving fitting configs
+    """
+
+    assert isinstance(fitting_configs, list)
+
+    if "train_file" not in use_params:
```
bernstei, line 193: For this wrapper to make sense in the context of the rest of wfl I think we should assume that the configs are received as a `ConfigSet`, and not support a parameter field that specifies a file. (see how `ace/fit.py` does it). Also, why are you initializing it to an empty dict? Later you set it to a string.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+    for key, val in mace_fit_params.items():
+        if isinstance(val, int) or isinstance(val, float):
+            mace_fit_cmd += f" --{key}={val}"
+        elif isinstance(val, str):
+            mace_fit_cmd += f" --{key}='{val}'"
+        elif val is None:
+            mace_fit_cmd += f" --{key}"
+        else:
+            mace_fit_cmd += f" --{key}='{val}'"
+
```
bernstei, line 136: please use double blank lines only between functions, not inside one.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+        else:
+            mace_fit_cmd += f" --{key}='{val}'"
+   
+ 
+    if not do_fit or verbose:
+        print('fitting command:\n', mace_fit_cmd)
+    
+    # Not totally sure if this part is applicable to MACE cpu training.   
+    orig_omp_n = os.environ.get('OMP_NUM_THREADS', None)
+    if 'WFL_GAP_FIT_OMP_NUM_THREADS' in os.environ:
```
bernstei, line 143: As I said above, if this is useful, change GAP to MACE.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+    
+        sys.stdout.write(stdout)
+        sys.stderr.write(stderr)
+    
+        # no outputs to rename since everything should be in run_dir
+        xpr.mark_processed()
+        
+        return results
+    
+    if not run_dir.exists():
```
bernstei, line 119: just use `exist_ok=True` instead of the `if`

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+        run_dir.mkdir(parents=True)
+    
+    if isinstance(mace_fit_params, str):
```
bernstei, line 122: I'd prefer to simplify this function by requiring that the params be passed to this function in the form of a dict.  If the calling function wants to load that dict from a yaml, that's fine, but isn't `fit()`'s business.

wfl/fit/mace.py
```diff
@@ -0,0 +1,205 @@
...
+        print('fitting command:\n', mace_fit_cmd)
+    
+    # Not totally sure if this part is applicable to MACE cpu training.   
+    orig_omp_n = os.environ.get('OMP_NUM_THREADS', None)
+    if 'WFL_GAP_FIT_OMP_NUM_THREADS' in os.environ:
+        os.environ['OMP_NUM_THREADS'] = os.environ['WFL_GAP_FIT_OMP_NUM_THREADS']
+    
+    try:
+        remote_cwd = os.getcwd()    
+        if str(run_dir) != ".":
```
bernstei, line 148: why does the run_dir value affect how this is handled?

wfl/fit/mace.py
jungsdao: In my simple test, when used with CPU, MACE seems to use multiple cores according to `OMP_NUM_THREADS`. so I retained that part with modified variable name.

wfl/generate/minimahopping.py
```diff
@@ -1,107 +1,134 @@
...
+            config_type_append(at, 'traj')
+            md_traj.append(at)
+
+    return md_traj
+
+
+# perform MinimaHopping on one ASE.atoms object
+def _atom_opt_hopping(atom, calculator, Ediff0, T0, minima_threshold, mdmin,
+                     fmax, timestep, totalsteps, skip_failures, **opt_kwargs):
+    fit_idx = opt_kwargs.pop("fit_idx", 0)
```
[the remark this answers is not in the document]
jungsdao: The reason I did not used `mkdtemp` was to keep track of minima hopping run. But It's true that it might not work when autoparallelized. so I reverted back to using `mkdtemp`

wfl/fit/mace.py
jungsdao: I changed `do_fit` to `dry_run` and it just prints mace fitting keywords.

wfl/fit/mace.py
jungsdao: Assuming two initialization situations: 1) List of ASE objects 2) xyz file. 
In case initiated with list of ASE object, it writes xyz file when fitting, but requires `fit_cfgs_fname` as argument. 
If initiated from xyz file, then it's copied to `run_dir`. 

Also, I wanted to have situation where user wants to restart from previous checkpoint file, then it should be copied by `remote_info.input_files` and `prev_checkpoint_file` argument should be provided.

bernstei: One of the points of `wfl` is the abstraction of input lists of configs. Please take a look at how the other fitting codes deal with the input coming in as a `ConfigSet`, which could be a file, or many files, or a list of configs, or in principle even something else (e.g. a query to a database that returns a particular set of configs you can iterate over).

bernstei: Note that checkpointing is a separate can of worms, because it isn't a list of configs that wfl has specialized infrastructure for. I'll take a look at the revised code.

wfl/generate/minimahopping.py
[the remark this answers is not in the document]
bernstei: Looks OK now, except that `fit_idx` is still defined and set to 0, even though it's never used.

wfl/fit/mace.py
```diff
@@ -0,0 +1,175 @@
...
+        will try to use env var WFL_EXPYRE_INFO used (see below). '_IGNORE' is for
+        internal use, to ensure that remotely running job does not itself attempt to spawn another
+        remotely running job.
+    remote_label: str, default None
+        label to match in WFL_EXPYRE_INFO
+    wait_for_results: bool, default True
+        wait for results of remotely executed job, otherwise return after starting job
+    prev_checkpoint_file: str, default None
+        Previous checkpoint file to restart from. 
+    fit_cfgs_fname: str, default None
```
bernstei, line 50: As I said elsewhere, I don't want to special case an xyz file for the fitting configs, since you can always pass a `ConfigSet` which wraps one file. The nice thing about doing it the wfl way (i.e. a ConfigSet) is that the contents will be automatically sent to a remote machine if needed, unlike a special file argument where that'd have to be dealt with manually.

wfl/fit/mace.py
```diff
@@ -0,0 +1,175 @@
...
+        If present and not None and not '_IGNORE', RemoteInfo or dict with kwargs for RemoteInfo
+        constructor which triggers running job in separately queued job on remote machine.  If None,
+        will try to use env var WFL_EXPYRE_INFO used (see below). '_IGNORE' is for
+        internal use, to ensure that remotely running job does not itself attempt to spawn another
+        remotely running job.
+    remote_label: str, default None
+        label to match in WFL_EXPYRE_INFO
+    wait_for_results: bool, default True
+        wait for results of remotely executed job, otherwise return after starting job
+    prev_checkpoint_file: str, default None
```
bernstei, line 48: Can you move this argument earlier, say before `skip_if_present`? I try to have the function-specific arguments first, then the sort of infrastructure ones (that control things like running remotely, etc), later.

wfl/fit/mace.py
jungsdao: I understood your concerns regarding `ConfigSet` and largely borrowed some parts from `GAP/simple.py`.

wfl/fit/mace.py
```diff
@@ -125,30 +133,29 @@ def fit(fitting_configs, mace_name, mace_fit_params, mace_fit_cmd, ref_property_
...
-        # If initiated with list of ASE objects then write xyz file in run_dir. 
-        if not fitting_configs.one_file():
-            fitting_configs = list(fitting_configs)
-            _write_fitting_configs(fitting_configs, mace_fit_params, run_dir, fit_cfgs_fname)
-        else:
-            copyfile(fitting_configs.one_file(), f"{run_dir}/{fitting_configs.one_file().stem}.xyz")
-
         # previous checkpoint file should be moved by remote_info.input_files function 
         if prev_checkpoint_file != None:
             checkpoint_dir = Path(run_dir / "checkpoints")
```
bernstei, line 138: Having the `Path( .. )` is useless here, because the `/` operator only works on `Path` objects to begin with.  If you're worried that `run_dir` isn't a `Path`, you need `Path(run_dir) / "checkpoints"`, but `run_dir` is used in so many places, you should just do `run_dir = Path(run_dir)` at the beginning of the function, if needed, and then you can assume it's a `Path` later in the function.

wfl/fit/mace.py
```diff
@@ -104,7 +103,16 @@ def fit(fitting_configs, mace_name, mace_fit_params, mace_fit_cmd, ref_property_
         return results
     
     run_dir.mkdir(parents=True, exist_ok=True)
-    
+   
+    # Copy training set xyz file to run_dir.
```
bernstei, line 107: I think it'd be nicer to have all of this logic in the same place, rather than having it split between here and the `_write_fitting_configs` function. For GAP it's all in this top level function, but in retrospect I think it's be better if it was all in a second smaller function, so if you do that here, I can refactor the GAP (and ACE?) functions. In fact, I almost think it's such a standard thing for fitting codes that we should write a single utility function that all fitting wrappers can call, but I can do that later.

wfl/fit/mace.py
bernstei: Looks good.  See my little remaining suggestion about refactoring that code.

bernstei:
@jungsdao looks like it's practically ready. A couple tiny changes left, and I'll enable the testing.  For that I'll need to make sure the CI has mace fitting installed, which I can do in a separate PR which you may need to merge from so the automated tests will run properly.  I'll let you know.

jungsdao:
> @jungsdao looks like it's practically ready. A couple tiny changes left, and I'll enable the testing. For that I'll need to make sure the CI has mace fitting installed, which I can do in a separate PR which you may need to merge from so the automated tests will run properly. I'll let you know.

I have applied modifications (e.g. refactoring functions) according to your suggestions. Hope everything works well by now.

wfl/fit/mace.py
bernstei: Please use python recommended `is not None` rather than `!= None`

wfl/fit/mace.py
bernstei: Can you simplify this bit of code to something like
```python
fitting_configs_scratch_filename = _prep_fitting_configs_file(mace_fit_params, fitting_configs)
```
rename `_write_fitting_configs` to `_prep_fitting_configs_file`, and add to it the call to `one_file()` and the `if ... else ...`. that are currently here? That way all the updating of `mace_fit_params` will happen in one place, and this modified function can return `None` if there's no scratch file, so the `unlink` later will still work.

bernstei:
OK - an unreasonable amount of fighting with it later, once #267 passes the tests I'll merge it to `main`, then you can merge `main` into this branch, so we can run the pytests in the github CI. I'll let you know when it's ready.

wfl/fit/mace.py
jungsdao: I think I have misunderstood your first comment, and I have removed that separate function. But now I understood your point. I'll rewrite a function named with that and push again.

bernstei:
OK, now that #267 has been merged into main, can you merge from main into this one, so we can get the automated CI testing to work? I guess we also need to settle on how to find `run_train.py`, which will require mace to decide if they're going to merge my PR that makes it an executable script in `~/.local/bin` as part of the pip install process.

bernstei:
Please merge from main, so the tests including mace fitting can run.

tests/test_mace_fitting.py
```diff
@@ -0,0 +1,56 @@
...
+mace_fit_cmd = f"python {script_dir}/scripts/run_train.py"
+
+
+def test_mace_fit_from_list(request, tmp_path, monkeypatch, mace_fit_cmd=mace_fit_cmd):
+
+    monkeypatch.chdir(tmp_path)
+    print('getting fitting data from ', request.fspath)
+    print("here :  ", Path(request.fspath))
+
+    params_file_path = Path(Path(request.fspath).parents[0] / 'assets' / 'mace_fit_parameters.yaml')
```
bernstei, line 24: Please remove this and all the other unnecessary `Path(Path(...)` calls.   The `/` operator returns a `Path` object, and `tmp_path` is guaranteed to be a `Path`, so you don't have to re-convert it repeatedly.

Also, you can just define
```python
parent_path = Path(request.fspath).parent
```
and use that everywhere in the function.

bernstei: Also, `request.fspath` has been deprected https://docs.pytest.org/en/latest/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path, although I haven't figured out the approved replacement yet.

bernstei: https://docs.pytest.org/en/7.4.x/reference/reference.html#std-fixture-request
Looks like it's `request.path`, which is already of type `Path`.

tests/test_mace_fitting.py
```diff
@@ -0,0 +1,58 @@
...
+
+    monkeypatch.chdir(tmp_path)
+    print('getting fitting data from ', request.path)
+    print("here :  ", request.path)
+
+    parent_path = request.path.parent
+    fit_config_file = parent_path / 'assets' / 'B_DFT_data.xyz'
+    params_file_path = parent_path / 'assets' / 'mace_fit_parameters.yaml'
+    mace_fit_params = yaml.safe_load(params_file_path.read_text())
+    mace_fit_params["train_file"] = fit_config_file
```
bernstei, line 48: It's going to ignore the `"train_file"` dict entry anyway, right? We should either test without it set, or with it set to the wrong thing, to test the overriding.

jungsdao: It's true that dict assignment is ignored anyway if `fitting_config` is initiated from an already writtne file. 
pytest works either way whether I set dict entry with "train_file" or not.

bernstei: Yes, but if someone breaks the ConfigSet single file writing code in the mace fit wrapper, and you pass the correct file, the pytest could still pass, even though the code that writes to a file is broken.

jungsdao: I don't understand the situation of breaking the ConfigSet single file. Could you elaborate a little bit on that situation?

bernstei: The point of the tests is to make sure the code is working. If you write a test that passes even if the code is broken, that's not a useful test.  The whole point of the way the wrapper is written is that it shouldn't depend on the user passing the `"train_file"` parameter, because it's going to write the input file (if needed) by itself. If you pass that parameter, it's more likely that some time in the future something about the code that writes to a single file if needed will be broken accidentally, but this test won't fail, because it's also using a mechanism that we don't want to pass the location of the training data file.

jungsdao: Okay, if I understood correctly, I should not assign "train_file" and pass it on to `mace_fit_params` dict in this pytest.
I'll remove that dict entry assignment.

bernstei:
@jungsdao I enabled running the CI tests. Once they pass we can merge, although depending on when the MACE update to the training script path is merged, I may wait and ask you to update the pytest to use the new path. Do you agree that the CI should be using the mace `develop` branch?

jungsdao:
> @jungsdao I enabled running the CI tests. Once they pass we can merge, although depending on when the MACE update to the training script path is merged, I may wait and ask you to update the pytest to use the new path. Do you agree that the CI should be using the mace `develop` branch?

I already have tried pytest with mace `develop` branch, but I've encountered following error. 
This seems to be trivial difference, but I think in current status, it might not pass `develop` branch but will pass `main` branch I guess. 
```
Traceback (most recent call last):
  File "/raven/u/hjung/Softwares/mace/scripts/run_train.py", line 558, in <module>
    main()
  File "/raven/u/hjung/Softwares/mace/scripts/run_train.py", line 50, in main
    collections, atomic_energies_dict = get_dataset_from_xyz(
  File "/u/hjung/conda-envs/mace_env/lib/python3.9/site-packages/mace/tools/scripts_utils.py", line 41, in get_dataset_from_xyz
    atomic_energies_dict, all_train_configs = data.load_from_xyz(
  File "/u/hjung/conda-envs/mace_env/lib/python3.9/site-packages/mace/data/utils.py", line 215, in load_from_xyz
    if isolated_atom_config:
UnboundLocalError: local variable 'isolated_atom_config' referenced before assignment
```

tests/test_mace_fitting.py
jungsdao: I've passed on wrong "train_file" to `mace_fit_params` but it still works well with one connected to `fitting_configs`.

bernstei:
Can you edit `.github/workflows/pytests.yml` and remove the `@develop` from
```
python3 -m pip install git+https://github.com/ACEsuit/mace.git@develop
```
so it just tests the default branch, which is `main`?

bernstei:
Looks like there's at least one code correctness issue that the tests identified, so please keep on fixing those until the tests pass, or ask for help if there's something you aren't sure how to fix.

bernstei:
Looks like the logic for finding `run_train.py` is just incompatible with where things are installed in the CI. That'll be easy to fix once mace merges my PR (although it's going into the develop branch for now, so who knows how long it'll be before it's in main), but maybe we should just add a `WFL_MACE_FIT_COMMAND` env var, analogous to how the ACE fitting uses `WFL_ACE_FIT_COMMAND` when the command isn't passed in as a function argument.

jungsdao:
> Looks like the logic for finding `run_train.py` is just incompatible with where things are installed in the CI. That'll be easy to fix once mace merges my PR (although it's going into the develop branch for now, so who knows how long it'll be before it's in main), but maybe we should just add a `WFL_MACE_FIT_COMMAND` env var, analogous to how the ACE fitting uses `WFL_ACE_FIT_COMMAND` when the command isn't passed in as a function argument.
> 
> [added] note that here it should be easier, because of not having to chase down julia, etc. Just use `WFL_MACE_FIT_COMMAND` if `mace_fit_cmd is None`.

There was an error in Langevin part in  `test_md.py`, but has been fixed. 
I can add `WFL_MACE_FIT_COMMAND` env var, but I'm not sure how should this variable be set in pytest. 
I guess finding `run_train.py` makes things very complicated. 
I wonder if it would be better to wait until mace `develop` branch is updated.

bernstei:
> There was an error in Langevin part in `test_md.py`, but has been fixed. 

good

> I can add `WFL_MACE_FIT_COMMAND` env var, but I'm not sure how should this variable be set in pytest. 

The script that the CI runs knows exactly where it cloned the repo into, so it knows exactly where the script is, so it can set it, and the pytest can just check if it's set. If not, it can search for `mace_run_train` in the path (`shutil.which()`, I think), and then skip the test if it can't find it in either place. 

something like
```
mace_fit_cmd = os.environ.get("WFL_MACE_FIT_COMMAND", shutil.which("mace_run_train"))
.
.
.
@pytest.mark.skipif(mace_fit_cmd is None, reason="No mace_run_train found in WFL_MACE_FIT_COMMAND or path")
def test_.....(...):
   .
   .
   fit(.. mace_fit_cmd=mace_fit_cmd)
```
> I guess finding `run_train.py` makes things very complicated. I wonder if it would be better to wait until mace `develop` branch is updated.

In fact, it's basically impossible in general, and probably in fact impossible the way the CI workflow is currently set up, because I don't explicitly clone the repo, just pip install straight from github, so there's probably no local copy of the script directory. I think we should do what I suggested above, just because it's a versatile design, but if we don't wait for the version of MACE that installs `mace_run_train` into `PATH`, I'll have to change how mace is installed in the github CI.

jungsdao:
> > There was an error in Langevin part in `test_md.py`, but has been fixed.
> 
> good
> 
> > I can add `WFL_MACE_FIT_COMMAND` env var, but I'm not sure how should this variable be set in pytest.
> 
> The script that the CI runs knows exactly where it cloned the repo into, so it knows exactly where the script is, so it can set it, and the pytest can just check if it's set. If not, it can search for `mace_run_train` in the path (`shutil.which()`, I think), and then skip the test if it can't find it in either place.
> 
> something like
> 
> ```
> @pytest.mark.skipif(os.environ.get("WFL_MACE_FIT_COMMAND", shutil.which("mace_run_train")) is None,
>                                    reason="No mace_run_train found in WFL_MACE_FIT_COMMAND or path")
> def test_.....(...):
>    .
>    .
>    fit(.. mace_fit_cmd=None)
> ```
> 
> and modify the mace fitting wrapper to search, only if `mace_fit_cmd is None`, using the same `os.environ.get... shutil.which` construction.
> 
> > I guess finding `run_train.py` makes things very complicated. I wonder if it would be better to wait until mace `develop` branch is updated.
> 
> In fact, it's basically impossible in general, and probably in fact impossible the way the CI workflow is currently set up, because I don't explicitly clone the repo, just pip install straight from github, so there's probably no local copy of the script directory. I think we should do what I suggested above, just because it's a versatile design, but if we don't wait for the version of MACE that installs `mace_run_train` into `PATH`, I'll have to change how mace is installed in the github CI.

I think `mace_run_train` cannot be traced with `which` command since there's no executable with that name. 
So I just have included skipping part if `WFL_MACE_FIT_COMMAND` is not found.

tests/test_mace_fitting.py
```diff
@@ -0,0 +1,54 @@
+import yaml, os, shutil
+import time
+from pathlib import Path
+import mace, json
+from wfl.fit.mace import fit
+from wfl.configset import ConfigSet
+import ase.io
+import pytest
+
+os.environ["WFL_MACE_FIT_COMMAND"] = "python /raven/u/hjung/Softwares/mace/scripts/run_train.py"
```
bernstei, line 10: Please remove things that are hard-wired to your specific local setup. If you need that env var set when you run your pytests, just set it in your shell.  When it runs pytest the github CI workflow will either set that env var or use a version of mace that has the executable in path.

tests/test_mace_fitting.py
```diff
@@ -0,0 +1,54 @@
...
+
+    t0 = time.time()
+    fit(fitting_configs, "test", mace_fit_params, mace_fit_cmd=None, run_dir=".")
+    t_run = time.time() - t0
+
+    assert (tmp_path / "test.model").exists()
+    assert (tmp_path / "test.model").stat().st_size > 0
+
+
+#@pytest.mark.skipif(not os.environ.get("WFL_MACE_FIT_COMMAND"), shutil.which("mace_run_train") is None, reason="No mace_run_train found in WFL_MACE_FIT_COMMAND or path")
```
bernstei, line 34: is there a reason you commented this one out?

tests/test_md.py
```diff
@@ -70,6 +71,23 @@ def test_NVT_const_T(cu_slab):
...
 
+def test_NVT_Langevin_const_T(cu_slab):
+
+    calc = EMT()
+
+    inputs = ConfigSet(cu_slab)
+    outputs = OutputSpec()
+
+    atoms_traj = md.md(inputs, outputs, calculator=calc, integrator="Langevin", steps=300, dt=1.0,
+                           temperature = 500.0, temperature_tau=100/fs)
```
bernstei, line 82: why are you dividing by fs here?

jungsdao: I wanted set friction 0.01. since it's multiplied with fs in `md/__init.py`, I divided here. 
```
174                 elif integrator == "Langevin":
175                     md_constructor = Langevin
176                     stage_kwargs["friction"] = 1 / (temperature_tau * fs)
```

tests/test_mace_fitting.py
jungsdao: This is my mistake. I was testing with commented one and some hard-wired stuffs. I'll fix this.

bernstei:
> I opened an issue, [ACEsuit/mace#197](https://github.com/ACEsuit/mace/issues/197), to address this problem. 

Fixed in https://github.com/ACEsuit/mace/commit/321d1b02ba05e1489c2c0665cdcf2ded64baace8 but you'll need to change the `config_type` info field to `"IsolatedAtom"` to be compatible with the `develop` branch. As far as I can tell that's the string they check for in `main` as well, so you'll need to update it regardless of which branch we test with.

bernstei:
[ACEsuite/mace#178](https://github.com/ACEsuit/mace/pull/178) is merged into `develop`, so searching for `mace_run_train` in the path should just work.  Do you want to switch back to the `develop` branch, and see if it all just works?

jungsdao:
> [ACEsuite/mace#178](https://github.com/ACEsuit/mace/pull/178) is merged into `develop`, so searching for `mace_run_train` in the path should just work. Do you want to switch back to the `develop` branch, and see if it all just works?

Now it can search `mace_run_train` without skpping the pytest.

bernstei:
> I have added `B_DFT_data_mace_ftting.xyz` file with `config_type` `IsolatedAtom`. This seems to be the easiest way to get around this issue.

It's the only correct way, unless we want to have the wrapper guess the label the user might have used to indicate the single atom configs.

bernstei:
If the tests pass I'm happy to merge, and later I may refactor the mace fitting executable finding behavior, as described in #268

bernstei:
@jungsdao looks like the tests are passing, and I'l like to take another look at the new mace ones, but assuming they're OK, are you OK with me merging this one?

jungsdao:
> @jungsdao looks like the tests are passing, and I'l like to take another look at the new mace ones, but assuming they're OK, are you OK with me merging this one?

I got a comment from my colleague that he wanted to explicitly specify `validation_file` and `test_file` in mace fitting but its location was not accessible when submitted to remote using expyre. Maybe should I fix this issue before merge?

bernstei:
> Maybe should I fix this issue before merge?

One of the advantages of using `ConfigSet` to send configs to a function is that wfl automatically makes sure those configs are available when running via expyre.  Are you just talking about adding another optional argument with a `ConfigSet` for the testing configurations?

bernstei:
> [edited] The `wfl` way would be to pass another `ConfigSet`, but I guess it's true that I'd have to think about how to make it available in the remote run with expyre. Let me think about that.

The way GAP and ACE fitting handle this is by making sure the configs are in memory (you'd have to do this separately for the fitting and testing `ConfigSet`s), then passing those lists to the function that's wrapped with expyre, then that lower level function writes them to (temporary) files. 

It's definitely the most robust way of doing things, although it does increase the overhead by generating temporary config files even if the fit is running on the same system as the top level wfl script.  You'd have to convince me that any alternative is worth it, because it'll be at least a little cumbersome, and I don't think the overhead is all that significant.

jungsdao:
> > [edited] The `wfl` way would be to pass another `ConfigSet`, but I guess it's true that I'd have to think about how to make it available in the remote run with expyre. Let me think about that.
> 
> The way GAP and ACE fitting handle this is by making sure the configs are in memory by explicitly turning them into a `list` (you'd have to do this separately for the fitting and testing `ConfigSet`s), then passing those lists to the function that's wrapped with expyre, then that lower level function writes them to (temporary) files.
> 
> It's definitely the most robust way of doing things, although it does increase the overhead by generating temporary config files even if the fit is running on the same system as the top level wfl script. You'd have to convince me that any alternative is worth it, because it'll be at least a little cumbersome, and I don't think the overhead is all that significant.

I've added two more arguments which take `ConfigSet` as input. Also some modifications on `_prep_fitting_configs_file` to take whatever `configs` with different keys. Hope this conforms to wfl way of this issue.

bernstei:
> I've added two more arguments which take `ConfigSet` as input. Also some modifications on `_prep_fitting_configs_file` to take whatever `configs` with different keys. Hope this conforms to wfl way of this issue.

I think we need to look more carefully at making this work for remote runs. It's inherently somewhat complicated, and the current code is still relying on filenames instead of treating `ConfigSet`s as opaque repositories for configs, which I think won't be properly staged out for remote expyre jobs.

bernstei:
@jungsdao why are there changes to the gap fitting wrappers? I never noticed them before.

bernstei:
OK - I'm going to run the tests again, and if they pass, I'll merge. I'll update the way remote mace fitting jobs are handled separately.

jungsdao:
> OK - I'm going to run the tests again, and if they pass, I'll merge. I'll update the way remote mace fitting jobs are handled separately.

Sorry, something was a bit messed up when merging before but it should've been resolved by now. 
Probably how things are working now is not the best option, but I'm not sure how remote behavior can be improved within my knowledge. 
I appreciate for the comments and help with this PR.

PR #282, opened 2024-01-17
wait_for_results and fix multiple copies of output configs when interrupted and restarted

### LINKED ISSUE #281: duplicate configs in output
I've now seen twice that DFT evaluation tasks end up duplicating some configurations in the output file.  Still working on reproducing this systematically, but creating this issue so we don't forget.

bernstei:
Add `wait_for_results` argument to normal autoparallelized ops (not just fitting) so many can be started without blocking.

Wipe tmp.* files when creating `OutputSpec` so current practice of opening them for append (mode "a") will not lead to multiple copies of initial configs if script is stopped mid-op and rerun.

closes #281

### DIFF
--- a/.github/workflows/pytests.yml
+++ b/.github/workflows/pytests.yml
@@ -178,7 +178,7 @@ jobs:
           echo "which pw.x"
           which pw.x
           ls -l /usr/bin/pw.x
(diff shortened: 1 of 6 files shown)


bernstei:
Also fixes github CI espresso tests (env var), and incompatibility of espresso calc with latest ASE version.

PR #292, opened 2024-02-07
DFT calculator convergence test

bernstei:
Change the way calculator results copying checks for DFT calculator convergence to be less Vasp specific. 
Add test for convergence status.


bernstei:
Also had to update other tests to work with the latest ASE version, since they made `EMT` unpickleable, and also much faster, so a test that relied on a job timing out started failing.

PR #305, opened 2024-05-16
Add yml file for pypi workflow, change dependencies to PyPI releases

hheenen:
Hi, this pull-request adds the yml file for the PyPI distribution continuously integrated via Trusted Publisher Management, publishing is invoked by tags. Also the `setup.py` is modified with general dependencies from pypi for univeralSOAP and ExPyRe. Without adding the pull-requests in these repos first, I guess the pip install will fail.


hheenen:
Should we just close this PR? I created a fresh one!

PR #311, opened 2024-05-23
Refactor code that stores calculator results and updates config_type during generate ops (md, optimize, neb).

### LINKED ISSUE #310: `config_type` get concatenated over different operations
It is common to do multiple operations like optimizations and NEB calculation on the same config. 
After trying a few operations in that way I found that `config_type` has weird value like this. 
` 'config_type': 'optimize_last_converged_optimize_last_converged_neb_last_converged'`

I think it does make sense to keep only the last operations rather than keeping the history of all the operations. 
Also, `info` & `arrays` dictionary have two properties like 
'optimize_energy', 'neb_energy'
'optimize_forces', 'neb_forces'
Which could give confusion in case if the final operation was NEB calculation, it's still containing unrelated information about "optimization". 
I can probably select which value is relevant, but at this stage, optimize_energy and optimize_forces are totally redundant. 
I just wanted to point out this issue.

bernstei:
Stored properties calculated during `wfl.generate` ops now (by default) overwrite previous operations. New scheme is that if `<some_string>__` is in the property prefix, overwrite all properties that start with the same string.  All ops in `generate` use `last_op__`, i.e. `last_op__md`, `last_op__optimize_...`, etc, so they will overwrite each other.  Other `results_prefix` values can be used to override this default.

Defaults to appending to `atoms.info["config_type"]` by setting `update_config_type="append"` argument to various generate ops.  Also support `update_config_type="overwrite"` to overwrite previous value, and `False` to keep value the same.

closes #310


bernstei:
Should these various keys include a `wfl_` prefix, to avoid conflicting with misc user-set properties?

PR #313, opened 2024-06-04
Aims profile update

gelzinyte:
Mainly clean up the `AimsProfile` construction to work with the latest ASE release.


wfl/calculators/aims.py
```diff
@@ -65,19 +65,13 @@ def __init__(self, keep_files="default", rundir_prefix="run_Aims_", workdir=None
...
-                except TypeError:
-                    binary, parallel_info = parse_genericfileio_profile_argv(argv)
-                    kwargs_command["profile"] = AimsProfile(binary=binary, parallel_info=parallel_info)
+
+        if AimsProfile is None:
+            # old syntax
+            warnings.warn("Support for ASE 3.22-style calculator interfase will soon be depreciated")
+            kwargs_command["aims_command"] = calculator_exec
+        elif "profile" not in kwargs_command:
+            kwargs_command["profile"] = construct_aims_profile(calculator_exec, kwargs_command)
```
gelzinyte, line 74: On a second though, I think wfl should expect people to provide `profile` themselves among the kwargs. So I'd leave the two `elif` lines and `construct_aims_profile` out.

bernstei:
Thanks.  Should we make sure this is done for all the other DFT calculators that support profile? Even better if we could somehow abstract out a common function to deal with this.  

Or we could go all in on the new ASE approach, only support profile. Unless we hate it enough to keep the workaround :)

gelzinyte:
I would rather have as few modifications on top of ASE as possible so that wfl is easier to pick up and behaves more as expected. Also, setting up the profile yourself isn't that big of a deal, once ASE documentation is updated, if it's pickleable (will check).  

Enforcing a profile wouldn't be backwards compatible, though and I think people will take a while to update their scripts to ASE3.23, since a lot changed. Should we
a) tag a latest ase-3.22-compatible version or
b) keep both interfaces with something like `if CalculatorProfile is not None: assert "profile" is in kwargs`?

Let me look through how other calculators behave.

wfl/calculators/aims.py
bernstei: looks good, except a couple of spelling things: "interface", and "deprecated".

wfl/calculators/utils.py
```diff
@@ -22,31 +22,5 @@ def clean_rundir(rundir, keep_files, default_keep_files, calculation_succeeded):
     else:
         clean_dir(rundir, keep_files, force=False)
 
-def parse_genericfileio_profile_argv(argv):
```
bernstei, line 25: What's the deal with this function?  Is it just not used?

bernstei (changes requested):

wfl/calculators/utils.py
gelzinyte: Yes, they simplified a lot what the "Profile" does. It used to parse for the mpi executable, num tasks, etc, now it's all in one string provided by the user.

gelzinyte: On a quick look, the parsing is now done in the BaseProfile

bernstei:
> I would rather have as few modifications on top of ASE as possible

I agree in principle. In that case, can we avoid doing anything with the keywords, and assume the user is passing the right ones (consistent with whatever ASE versions they have installed)?  Or does that conflict with, e.g. the fact that we want each  job to run in a separate subdirectory (e.g. for VASP)?

bernstei:
> I would rather have as few modifications on top of ASE as possible so that wfl is easier to pick up and behaves more as expected. Also, setting up the profile yourself isn't that big of a deal, once ASE documentation is updated, if it's pickleable (will check).

What do we think about doing _nothing_ to the keywords, then, and leaving it up to people to pass the keywords that make sense for their ASE version?  Is that feasible, or does the fact that we want to run each job in a separate subdirectory make that impossible?

wfl/calculators/aims.py
gelzinyte: Fixed!

gelzinyte:
I've moved the general non-aims discussion to #314. 

For aims, we can drop messing with the execution-related kwargs. Then the user should give
 - 3.22: "aims_command" (e.g. `srun -n N path/to/aims_bin.x `) 
 - 3.23 profile (which gets loaded automatically from ase config file or the user can give a profile=AimsProfile(aims_command=`srun -n N path/to/aims_bin.x `) among the kwargs)

`_setup_calc_params` is untouched. 

To Do 

- [ ] Check `AimsProfile` is pickleable
- [ ] Update docs examples
- [ ] Update unit tests

bernstei:
@gelzinyte I know you preferred separate PRs, but I'm worried that since the testing is now happening on the latest ASE, various unrelated tests will fail, and it'll be a bit tricky to figure out whether the failures that are left in any given PR are or are not related to that PR, or to the other calculator wrappers that have't been updated yet.

Can you tell for this PR which ones are meant to be fixed, and which ones we can expect to continue to fail until we update the other calculators?  Also, do we want to keep two versions of the _tests_, ones that use `command` and ones that use a `profile`, depending on what ASE version it detects?

gelzinyte:
I mainly suggested separate PRs because I thought it'd take me longer to test the calculators I have never used and would need to install/compile. Separate PRs would also make it easier for others to contribute meanwhile. 

Regardless of the number of PRs, maybe I can first fix Aims, set CI tests to use pip-installed ase-3.23 and mark the not yet updated calculators' tests as x-failing for now. And go from there with either adding more calculators to this PR or merging to `main`. 

Having looked a bit more into the update, I would only keep the local ase configuration (profile)-based testing. They're only ever done locally, so the ase configuration file makes that convenient and it's for ase to make sure that both the configuration file and `profile` are supported. As for ase version-specific tests, I'm not sure if it's worth the effort to try and also support the 3.22 version.. (so no testing for `command`)

bernstei:
Looks like the VASP calculator doesn't have a profile yet.  Should I make a corresponding PR for VASP that we do at the same time, just to make sure that we don't introduce anything that completely breaks the older-style calculators, as long as those are still around?

Also, we could consider doing the same for Espresso, just to have a second data point for the ones that do use Profile.

gelzinyte:
> Looks like the VASP calculator doesn't have a profile yet. Should I make a corresponding PR for VASP that we do at the same time, just to make sure that we don't introduce anything that completely breaks the older-style calculators, as long as those are still around?

What would go in the PR, you mentioned some small changes? I haven't touched anything VASP-related and the tests on the github CI pass. Do your local VASP tests pass for this branch? If your worry is about more extensive testing with other calculators before this branch is merged, add them here? 

> Also, we could consider doing the same for Espresso, just to have a second data point for the ones that do use Profile.

If I were to update other calculators, I would rather come back to them at some later time. So a separate PR for QE would not hold up this branch? Or we could ask @jungsdao to contribute and test his changes directly here?

bernstei:
> What would go in the PR, you mentioned some small changes?

Just removing all the special handling of kwargs, basically, and then making sure that the tests are updated.  But I guess as long as this PR passes by itself, there isn't really a need for the Vasp one to be tested at the same time.

.github/workflows/pytests.yml
```diff
@@ -42,9 +42,14 @@ jobs:
 
       - name: Install latest ASE from gitlab
         run: |
-          python3 -m pip install git+https://gitlab.com/ase/ase.git
+          python3 -m pip install ase
```
bernstei, line 45: Should we specify a version, now that it might start mattering? `ase==3.23.0`, I guess?

docs/source/first_example.md
```diff
@@ -82,7 +82,7 @@ from ase.calculators.emt import EMT
 from wfl.calculators import generic
 from wfl.autoparallelize import AutoparaInfo
 from wfl.configset import ConfigSet, OutputSpec
-from wfl.autoparallelize.remote info import RemoteInfo()
+from wfl.autoparallelize.remote.info import RemoteInfo()
```
bernstei, line 85: imports just need names, not function calls (i.e. no `()`)

bernstei:
> Just removing all the special handling of kwargs, basically, and then making sure that the tests are updated. But I guess as long as this PR passes by itself, there isn't really a need for the Vasp one to be tested at the same time.

I knew I had a reason to be worried.  VASP has separate executables for gamma-point only calculations, and now we're dealing with that, I don't know how to generalize the current mechanism, which is to accept `command_gamma` and `calculator_exec_gamma` arguments, save them someplace special, and modify the underlying ASE calculator when they are needed. I'll try to simplify the current approach, but I don't really know how it's going to work if/when Vasp starts using a profile.

docs/source/first_example.md
gelzinyte: And even `remote.info` isn't correct.. Fixed this and other on the same page.

.github/workflows/pytests.yml
gelzinyte: I initially thought that it might be good to keep up with the latest pip release, but it might be nicer to choose when to update manually. Added the version (`v3.23.0`)

gelzinyte:
I think it's fine to leave VASP (CASTEP, ..) similar to how it's been so far until it's clear how it's updated in ASE?

.github/workflows/pytests.yml
bernstei: Actually, it might be interesting to have to CI jobs, one that needs to pass, which tests on a version of ASE we are committing to support, and another on the latest gitlab HEAD, which is just informational for us.  But not part of this PR.

bernstei:
> I think it's fine to leave VASP (CASTEP, ..) similar to how it's been so far until it's clear how it's updated in ASE?

I think we should clean up all the `command_exec` stuff, but I think that'll be quite easy.  Let's move this to the PRs for the individual calculators, like I created for Vasp #317

bernstei:
@gelzinyte please let me know explicitly when you think it's ready for another (final?) review.

gelzinyte:
This should be it!

local_pytest.wst
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
bernstei, line 1: What is this file for?

gelzinyte: I was planning to have all of the local tests set up in one place, but only have aims so far. Similarly to `complete_pytest.tin`. Maybe somewhere in `tests` would be more appropriate?

bernstei: Can you give it an even more explicit name? I agree that `tin` isn't the best (that's the name for our cluster), but `workstation` is even more generic. Either use the machine you plan for it to be run on, or, if it's really specific to your individual setup (i.e. you had to install things yourself and they're in your bin directory), your name?

gelzinyte: Adding my name would probably be the most appropriate here (done).

bernstei: Why is the directory named "complete"? If you just mean the calculators where the simplification is complete, maybe put that in the _script_ name? Are you envisioning other things that would belong in the "complete" subdir?  Also, if that file is meant to be kept long term, maybe we should make a `tests/local_scripts` subdirectory, and move the `.tin` and your script into there.

bernstei: Looks good.  Should I merge if I'm still happy with everything?

gelzinyte: Please do!

docs/source/examples.index.md
```diff
@@ -7,7 +7,6 @@ Examples often show examples of more than one thing. Below is a list of common o
 ## Evaluate structures with a calculator
  
 - ORCA: [ORCA via python script](examples.orca_python.md)
-- MACE: [Parallelize MACE calculator](examples.mace.md)
```
bernstei, line 10: Just noticed that you removed the MACE example.  Why?

wfl/calculators/aims.py
```diff
@@ -3,6 +3,7 @@
 """
 
 import shlex
+import warnings
```
bernstei, line 6: this import is unused, as far as I can tell

docs/source/examples.index.md
gelzinyte: I have added it a long time ago and haven't kept it up to date; there's nothing special about parallelising MACE; MACE has its own documentation that will be more up-to-date. Though I maybe should have left it alone, on a better look there's not that much of a difference compared to the MACE documentation.

bernstei: OK, that's fine.

bernstei:
Given that this PR now specifies a particular ASE version in the CI testing, what do you think about changing the dependencies and/or docs (in particular the top level `README.md`) to indicate what we are now doing w.r.t. this dependencies and its version?

In fact, what _should_ we do about the ASE version in the installation file?  `>= 3.23.0`, but say that it's only tested against 3.23.0 itself?  I don't want to always force people to downgrade their ASE.

bernstei:
> Yes, at least a note in the README will be good.

Can you add that as part of this PR?
 
> Actually, why not to test against the latest pip-installable ase version? (That's why I initially didn't specify a version.) The latest wfl version should work with ase v3.22 with the correct keywords (even if we don't test this), so I think we can leave 'ase>=3.22.1' in the requirements.

I'm ambivalent about this.  Maybe we should discuss on the slack?

> 
> We should tag a new version of wfl (0.4.0?) once the other calculators are updated. We could then enforce ase 3.23, because there then would be an option to `pip install wfl==0.3`.

Yes, that's a good idea.

bernstei:
@gelzinyte why don't you remove the ase version from the CI, but leave it with pip, not github. And we'll leave the pyproject dependency as is.

gelzinyte:
Done. Have also updated readme and home page of documentation.

README.md
```diff
@@ -8,10 +8,12 @@ The main functions of Workflow is to efficiently parallelise operations over a s
...
 
-NOTE: because of the very large time intervals between official ASE releases, `wfl` is typically
-set up for (and tested against) the latest ASE gitlab repo `master` branch.  Recent changes
-that require this support include variable cell minimization using `FrechetCellFilter` and
-`Espresso` calculator configuration. See documentation link above for installation instructions.
+`wfl` its dependendensies may be installed vi `pip install wfl`. 
+
+NOTE: `wfl` is only tested against the latest ASE pip release, currently v3.23. 
+For the time being, ASEv3.22 is allowed with the most recent branch of `wfl`, hoewever,
+not fully supported (e.g. the correct file-based calculator keywords must be specified
```
bernstei, line 15: I don't think this is really an example of "not fully supported". It's got less automatic managing of those keywords, but it's perfectly happy to pass the args and/or env var to the ASE calculator.

I'm going to edit these sentences,

bernstei: If you are happy with the little wording changes I put it, please post here, and I'll merge.  If you think I mis-stated the current compatibility status, we should fix that first.

gelzinyte: That's clearer, thanks. 

(I've just fixed minor typos in "`wfl` its dependendensies may be installed vi `pip install wfl`. ")

README.md
```diff
@@ -8,7 +8,7 @@ The main functions of Workflow is to efficiently parallelise operations over a s
 
 For examples and more information see [documentation](https://libatoms.github.io/workflow/)
 
-`wfl` its dependendensies may be installed vi `pip install wfl`. 
+`wfl` and its dependendensies may be installed via `pip install wfl`.
```
bernstei, line 11: I may change this :) I'm on a campaign to end the `pip` command (in favor of `python3 -m pip ...`.

bernstei: > What's wrong with pip (if one makes sure it's the correct executable..)?

That's exactly what's wrong.  Especially in systems where there's a `python`(2) and `python3`, it can be `pip3`, etc. Admittedly even assuming `python3` is the right executable is not entirely safe, but I think it's safer.

bernstei (approved):

PR #317, opened 2024-06-11
simplify Vasp wrapper to minimize mangling to ASE Vasp calculator keywords

bernstei:
Initial pass removes `calculator_exec` and `_gamma`, but has to leave `command_gamma`.  

Also slightly generalize situations under which it will use the gamma-point-only executable to include KGAMMA=True and KSPACING large enough (rather than just pbc = False).


bernstei:
@gelzinyte do you have any comments, or should I just merge?

gelzinyte:
Could you check and update the documentation? I think VASP is only mentioned in `overview.queued.md` .  Looks good otherwise!

bernstei:
> Could you check and update the documentation? I think VASP is only mentioned in `overview.queued.md` . Looks good otherwise!

Pretty sure it's still correct, but I changed the env var to `ASE_VASP_COMMAND`, which I actually test (locally, of course, not in the CI).  Frankly, though, I'm using whole `RemoteInfo` dict in a an env var less these days, so maybe that example is due for a more major revision (but just as a matter of best practices, not in terms of being correct).  I think I'll leave it for a different PR, though. Perhaps coupled with dropping support for my complex system for deciding which json dict to use for which function to be parallelized.

PR #321, opened 2024-07-03
Quantum espresso wrapper for new ASE version

jungsdao:
Some modifications on QE wfl wrapper and slight modification on minima hopping as it doesn't save energy and force etc in info/arrays.

### DIFF
--- a/tests/calculators/test_qe.py
+++ b/tests/calculators/test_qe.py
@@ -11,6 +11,7 @@ import numpy as np
 import requests
 from ase import Atoms
 from ase.build import bulk
+from ase.calculators.espresso import EspressoProfile
 from pytest import approx, fixture, raises, skip
 
 # from wfl.calculators.espresso import evaluate_autopara_wrappable, qe_kpoints_and_kwargs
@@ -61,14 +62,16 @@ def test_qe_kpoints(tmp_path, qe_cmd_and_pseudo):
 
     qe_cmd, pspot = qe_cmd_and_pseudo
 
+    print("hi : ", qe_cmd, os.path.dirname(pspot))
     kw = dict(
         pseudopotentials=dict(Si=os.path.basename(pspot)),
         input_data={"SYSTEM": {"ecutwfc": 40, "input_dft": "LDA",}},
         kpts=(2, 3, 4),
         conv_thr=0.0001,
-        calculator_exec=qe_cmd,
+#        calculator_exec=qe_cmd,
         pseudo_dir=os.path.dirname(pspot),
-        workdir=tmp_path
+        workdir=tmp_path,
+		profile=EspressoProfile(qe_cmd, os.path.dirname(pspot))
     ) 
 
     # PBC = TTT
@@ -169,9 +172,10 @@ def test_qe_calculation(tmp_path, qe_cmd_and_pseudo):
         input_data={"SYSTEM": {"ecutwfc": 40, "input_dft": "LDA",}},
         kpts=(2, 2, 2),
         conv_thr=0.0001,
-        calculator_exec=qe_cmd,
+#        calculator_exec=qe_cmd,
         pseudo_dir=os.path.dirname(pspot),
-        workdir=tmp_path
+        workdir=tmp_path,
+        profile=EspressoProfile(qe_cmd, os.path.dirname(pspot))
     ) 
 
     calc = (wfl.calculators.espresso.Espresso, [], kw)
@@ -225,8 +229,9 @@ def test_wfl_Espresso_calc(tmp_path, qe_cmd_and_pseudo):
         input_data={"SYSTEM": {"ecutwfc": 40, "input_dft": "LDA",}},
         kpts=(2, 2, 2),
         conv_thr=0.0001,
-        calculator_exec=qe_cmd,
-        pseudo_dir=os.path.dirname(pspot)
+#        calculator_exec=qe_cmd,
+        pseudo_dir=os.path.dirname(pspot),
+		profile=EspressoProfile(qe_cmd, os.path.dirname(pspot))
     ) 
 
     calc = wfl.calculators.espresso.Espresso(
@@ -249,9 +254,10 @@ def test_wfl_Espresso_calc_via_generic(tmp_path, qe_cmd_and_pseudo):
         input_data={"SYSTEM": {"ecutwfc": 40, "input_dft": "LDA",}},
         kpts=(2, 2, 2),
         conv_thr=0.0001,
-        calculator_exec=qe_cmd,
+#        calculator_exec=qe_cmd,
         pseudo_dir=os.path.dirname(pspot),
-        workdir=tmp_path
+        workdir=tmp_path,
+		profile=EspressoProfile(qe_cmd, os.path.dirname(pspot))
     ) 
 
     calc = (wfl.calculators.espresso.Espresso, [], kw)
--- a/tests/calculators/test_qe.py
+++ b/tests/calculators/test_qe.py
@@ -5,6 +5,7 @@ import os
 from shutil import which, copy as shutil_copy
 from pathlib import Path
 import pytest
+from packaging.version import Version
 
 import ase.io
 import numpy as np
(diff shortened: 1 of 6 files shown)


wfl/calculators/espresso.py
```diff
@@ -59,57 +51,11 @@ class Espresso(WFLFileIOCalculator, ASE_Espresso):
...
-
-            if calculator_exec is not None:
-                if "command" in kwargs_command:
-                    raise ValueError("Cannot specify both command and calc_exec")
-
-                kwargs_command["command"] = f"{calculator_exec} -in PREFIX.pwi > PREFIX.pwo"
-
-            # command or env var must be set
-            assert "command" in kwargs_command or "ASE_ESPRESSO_CALCULATOR" in os.environ
+        assert "profile" in kwargs_command
```
bernstei, line 58: Please don't require the new version.  Just let the user pass whatever arguments they want, and if it's compatible with their ASE version, it'll work, if not, not.

wfl/calculators/espresso.py
```diff
@@ -140,9 +86,9 @@ def calculate(self, atoms=None, properties=_default_properties, system_changes=a
             calculation_succeeded = True
             if 'DFT_FAILED_ESPRESSO' in atoms.info:
                 del atoms.info['DFT_FAILED_ESPRESSO']
-            if "__calculator_output_prefix" in atoms.info:
-                save_calc_results(atoms, prefix=atoms.info["__calculator_output_prefix"], properties=properties)
-                atoms.info["__calculator_results_saved"] = True
+            if "_output_prefix" in atoms.info:
```
bernstei, line 89: Please undo this change.  If the `espresso` calculator modifies the configurations and breaks the ASE caching (the way the Vasp calculator does), you do need to save the results, but you need to use the property that's set by  `calculator.generic`, which is the previous version of this bit of code, i.e. `__calculator_output_prefix`. 

I actually suspect that this calculator doesn't modify the configurations, in which case the ASE caching should work, and you don't need to save the results explicitly here at all, but maybe this PR isn't the right place to make that change.

bernstei:
The tests are failing because `profile` is used in the tests but it is undefined. @gelzinyte can you take a look and do something analogous to what you did for aims? I don't exactly understand the whole new `Profile` business, and how you're calling Aims without passing `AimsProfile` object. Or do you want to closet this one and do it from scratch for espresso separately?

jungsdao:
In my local machine, profile is defined by ase_config file. I don't know how this config file should be defined for pytest. 
Also in my pytest in local machine, it actually fails because the `workdir=tmp_path` is defined as, for example, `tmp_path = PosixPath('/tmp/pytest-of-hjung/pytest-2/test_qe_calculation0')`, but actual workdir is `/tmp/pytest-of-hjung/pytest-2/test_qe_calculation0/run_QE_9f6rsnhc`. I'm not sure if this is expected behavior.

bernstei:
I think I have a fix for the testing profile issue.  Please apply the attached patch **before** you fix my comments above.  

Also, there's no good reason for the pytest to fail for you.  The `workdir` is supposed to be set to the directory in which the espresso calculator will be creating each `run_QE_xxxxxx` subdirectory, one for each config, and it will set `self.directory` to that, so the calculator should run in the right place. It certainly works for me, and when the github CI runs it.

[new_espresso_test.patch](https://github.com/user-attachments/files/16135711/new_espresso_test.patch)

tests/calculators/test_qe.py
```diff
@@ -55,20 +68,21 @@ def qe_cmd_and_pseudo(tmp_path_factory):
     pspot_file = tmp_path_factory.getbasetemp() / "Si.UPF"
     shutil_copy(Path(__file__).parent / ".." / "assets" / "QE" / "Si.pz-vbc.UPF", pspot_file)
 
-    return cmd, pspot_file
+    return EspressoProfile(command=pytest_command, pseudo_dir=pspot_file.parent), pspot_file.name
```
gelzinyte, line 71: Once the tests are shifted to rely on ase config, there's no need to return an `EspressoProfile` (it should be constructed automatically from the ase config file), we only need to specify the `pspot_file.name`. I would move the setting up of the pseudo dir to the github pytest setup and only have this function moving the `Si.UPF` to the `pseudo_dir` location, specified by the `espresso` in ase config file. (Then the `kw` would not need a `profile` at all in the tests below)

tests/calculators/test_qe.py
```diff
@@ -215,18 +228,18 @@ def test_qe_calculation(tmp_path, qe_cmd_and_pseudo):
     assert si2.arrays["QE_forces"][0] == approx(-1 * si2.arrays["QE_forces"][1])
 
 
-def test_wfl_Espresso_calc(tmp_path, qe_cmd_and_pseudo):
-    # command and pspot
-    qe_cmd, pspot = qe_cmd_and_pseudo
+def test_wfl_Espresso_calc(tmp_path, qe_profile_and_pseudo):
```
gelzinyte, line 231: Once the tests are shifted to rely on ase config and not on explicit profile construction, I would remove this test and only keep the`test_wfl_Espresso_calc_via_generic` below.

gelzinyte (commented):
Some comments on how I think we should test the "profile" calculators. I can take over finishing this PR @jungsdao ?

gelzinyte:
And the tests seem to fail because of phono3py, also on the main branch. Let me see about fixing that.

jungsdao:
> And the tests seem to fail because of phono3py, also on the main branch. Let me see about fixing that.

Hi @gelzinyte, I'm a bit stuck with phono3py part of pytest because I had no idea how why this was happening. 
Feel free to finish this PR, sorry for not finishing it by myself.

tests/calculators/test_qe.py
```diff
@@ -19,9 +21,24 @@
...
 from wfl.autoparallelize import AutoparaInfo
 
+if Version(ase.__version__) < Version("3.23"):
+     pytest.skip(reason="Quantum espresso tests are only supported for ASE v3.23, please update.")
+
+
+from ase.config import cfg as ase_cfg
+from ase.calculators.espresso import EspressoProfile
+
+pytest_command = os.environ.get("PYTEST_WFL_ASE_ESPRESSO_COMMAND")
```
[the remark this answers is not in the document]
bernstei: Is this before or after the patch I uploaded above? https://github.com/libAtoms/workflow/pull/321#issuecomment-2215901794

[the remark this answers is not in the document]
gelzinyte: This is after after (the env variable got introduced there, no?)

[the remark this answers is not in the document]
bernstei: That is effectively what happens with my patch, I believe, although it still needlessly defines and tests for that env var.  @gelzinyte if you create a PR I can edit based on this one + my patch, I can clean that up further.

tests/calculators/test_qe.py
bernstei: I wanted to test both automatically reading the profile from the config file and explicitly passing it to the wrapper.

tests/calculators/test_qe.py
bernstei: See comment above

bernstei:
> > And the tests seem to fail because of phono3py, also on the main branch. Let me see about fixing that.
> 
> Hi @gelzinyte, I'm a bit stuck with phono3py part of pytest because I had no idea how why this was happening. Feel free to finish this PR, sorry for not finishing it by myself.

See https://github.com/libAtoms/workflow/pull/153 for why we're doing it this way now

bernstei:
> > > And the tests seem to fail because of phono3py, also on the main branch. Let me see about fixing that.
> > 
> > 
> > Hi @gelzinyte, I'm a bit stuck with phono3py part of pytest because I had no idea how why this was happening. Feel free to finish this PR, sorry for not finishing it by myself.
> 
> See #153 for why we're doing it this way now

Looks like we can probably do it (i.e. install from source rather than trying to fight with conda) with `python3 -m pip install .` instead.

tests/calculators/test_qe.py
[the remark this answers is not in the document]
bernstei: OK, you're right.  It is still using it, although it doesn't actually need to.  If you create a PR I can edit, I'll fix that.

tests/calculators/test_qe.py
bernstei: The thing that makes it all tricky is that the tests need to use their own pseudopotentials, but the profile includes information on where the pseudopotential directory is.  So you _can't_ just rely on the user's generic profile (unless you copy the testing PPs into their directory).  I'll try to document this in the files a bit more carefully

gelzinyte:
Let me open a PR for phonopy, clone this PR and continue the discussion there?

bernstei:
> Let me open a PR for phonopy, clone this PR and continue the discussion there?

Sounds good.

wfl/calculators/espresso.py
gelzinyte: Removed, see #325

wfl/calculators/espresso.py
gelzinyte: Undid the change in #325 . Maybe let's see in #325 the calculator modifies the configs and remove if not.

PR #325, opened 2024-07-22
Update Quantum Espresso calculator

gelzinyte:


wfl/calculators/espresso.py
```diff
@@ -140,9 +84,6 @@ def calculate(self, atoms=None, properties=_default_properties, system_changes=a
             calculation_succeeded = True
             if 'DFT_FAILED_ESPRESSO' in atoms.info:
                 del atoms.info['DFT_FAILED_ESPRESSO']
-            if "__calculator_output_prefix" in atoms.info:
```
gelzinyte, line 143: The calculator only modifies the calculator kwargs, not configs themselves, so I have removed this check. Or should we keep it for completeness/future uses?

bernstei: I'm ambivalent. On the one hand, it isn't necessary, as you noted.  On the other hand, it doesn't do any harm, and I wonder if we should just leave it in there in case some later version _does_ change the configs (like the Vasp calculator does).

gelzinyte:
Continuing conversations from #321:

> I wanted to test both automatically reading the profile from the config file and explicitly passing it to the wrapper.

In that case, how about having one test that relies on the environment variable and the rest be local-profile-based?


> The thing that makes it all tricky is that the tests need to use their own pseudopotentials, but the profile includes information on where the pseudopotential directory is. So you can't just rely on the user's generic profile (unless you copy the testing PPs into their directory). I'll try to document this in the files a bit more carefully

Maybe that's uncceptable, but can we copy the test-pseudopotentials to where the profile points and name it something explicit, e.g. `wfl_pytest.Si.pz-vbc.UPF`? Also explicitly document this behaviour. 

Realistically, it would probably only be developers testing QE locally and I don't mind a file copied (and then deleted?) to where ase config points, because I have a separate just for ase testing.

wfl/calculators/espresso.py
gelzinyte: Let's keep this bit of code. I've undid my previous removal.

bernstei:
Is https://github.com/libAtoms/workflow/pull/325#issuecomment-2243311439 still an issue?

I settled on always using the user's default profile (no longer testing passing `profile` explicitly, since that's really an internal ASE issue), and taking advantage of the fact that you can pass `pseudo_dir` as an explicit argument to the calculator constructor, which overrides the profile, to make sure it uses the test's PP.

gelzinyte:
>Is https://github.com/libAtoms/workflow/pull/325#issuecomment-2243311439 still an issue? 

No, I think this is all ready to merge, once the the tests pass. Thanks!

bernstei:
> > Is [#325 (comment)](https://github.com/libAtoms/workflow/pull/325#issuecomment-2243311439) still an issue?
> 
> No, I think this is all ready to merge, once the the tests pass. Thanks!

I'll let you merge in the recent merged PR and we'll see what happens here.

bernstei:
I'm patching up the test failures, but do we even understand how this branch ended up with minima hopping tests that don't quite match the minima-hopping code?

bernstei:
Looks like the CI still isn't seeing the ASE config to get the profile from.  Working on it.

gelzinyte:
> I'm patching up the test failures, but do we even understand how this branch ended up with minima hopping tests that don't quite match the minima-hopping code?

@jungsdao has also contributed some minima hopping modifications in #321, which moved over with the branch. @bernstei , @jungsdao are the minima hopping changes also ok to merge to main?

jungsdao:
It looks good to me :) thanks!

bernstei:
Are we ready to merge?

PR #330, opened 2024-08-05
Castep ase323 update

gelzinyte:


wfl/calculators/castep.py
```diff
@@ -83,7 +90,7 @@ def calculate(self, atoms=None, properties=_default_properties, system_changes=a
 
         orig_pbc = self.atoms.pbc.copy()
         try:
-            super().calculate(atoms=atoms)
+            super().calculate(atoms=atoms, properties=properties, system_changes=system_changes)
```
gelzinyte, line 93: `properties` and `system_changes` required in ASE v3.23, not defined in `ase.castep.calculate()` in ASE v3.22.

wfl/calculators/castep.py
```diff
@@ -95,17 +102,18 @@ def calculate(self, atoms=None, properties=_default_properties, system_changes=a
             # ASE castep calculator does not ever raise an exception when
             # it fails.  Instead, you get things like stress being None,
             # which lead to TypeError when save_calc_results calls get_stress().
-            for property in properties:
-                result = self.get_property(property)
+            for prop in properties:
+                result = self.get_property(prop, allow_calculation=False)
```
gelzinyte, line 106: Reason for the recursion - in `get_property` checks if results are present and if not, triggers a calculation. `allow_calculation` stops from running the calculation, but it's a dummy/unused argument in ASE v3.22, so there's no way of preventing the calculation via this argument.

gelzinyte:
The ASE v3.22 and v3.23 Castep versions aren't easy to reconcile like it was in the other profile-based calculators. Mainly because they made it more consistent with the rest of the ASE calculators. I've left comments in the two relevant places. 

I think the most straightforward solution is to enforce ASE v3.23 for the castep implementation, while we're still allowing (if not testing) ASE v3.22 for the other calculators. 

To make both acceptable, we would need to check the version and implement different choices in the two places I've highlighted. For example, call `self.calculate(atoms) ` for v3.22 and `self.calculate(atoms, properties, allow_calculation)` for v2.23; check that `self.get_property()` returns `None` in v3.23 and check if self._energy_total` is `None` in v3.22. I think that's messy and we should move towards only allowing 3.23 in the requirements eventually. 

@bernstei thoughts?

There are no major changes otherwise and the tests pass with Castep23 locally, so this PR is ready to merge.

bernstei:
I want to do a brief review, but I think it'll be fine.

> we should move towards only allowing 3.23 in the requirements eventually.

My preference would be to just force the move to 3.23 now, because I'd rather keep our code simpler. And 3.23 was so delayed, it has tons of updates/fixes.  But I do admit that the changes were so substantial that I can see people wanting to stick for 3.22 for a while to avoid breaking existing scripts.

Is there any way for us to gauge whether any users actually want pre-3.23 support?

wfl/calculators/castep.py
bernstei: I take it that despite the CASTEP calculator being made more consistent with the rest of the ASE calculators, it still has the behavior that caused the need for this bit of code (i.e. `None` properties instead of raising an exception for failures)?

bernstei:
See the one brief question/comment above, otherwise let's decide about the 3.22 support, and then merge.

wfl/calculators/castep.py
gelzinyte: I don't know which errors exactly weren't being raised.  Looking at the code, some warnings in 3.22 have been promoted to exceptions in 3.23, so we could assume that this issue has been fixed. I'd lean towards keeping this, though. It's a cheap check and should catch all the failures that still slipped through the update.

gelzinyte:
> My preference would be to just force the move to 3.23 now, because I'd rather keep our code simpler. And 3.23 was so delayed, it has tons of updates/fixes. But I do admit that the changes were so substantial that I can see people wanting to stick for 3.22 for a while to avoid breaking existing scripts.

> Is there any way for us to gauge whether any users actually want pre-3.23 support?

In that case, I would like to update the requirement to 3.23. I think that for 3.22 compatibility, we could add a note in the README, pointing to the latest wfl version before this update/next tag. I'll add the relevant changes (in this PR?) if that sounds good.

wfl/calculators/castep.py
bernstei: sounds good

gelzinyte:
Maybe let's first merge this and orca PRs and bump/fix the version & pip in a separate PR (I'm a bit confused about how it works, so would like to not mix it in here).

PR #333, opened 2024-08-15
Update ORCA to work with ASE v3.23

gelzinyte:
Some notes on the changes: 
- Update to work with the ASE's "profile" update
- The old code was re-implementing `ase.orca.ORCA.calculate`. This PR  instead calls the parent `self.calculate()` and all the modifications are moved in before and after. These are
    - Fill in any missing parameters with the default values. (The default values are unfortunately copy-pased from ase's orca.py, so can fall out of date, but ase does not expose them.) ASE does that [here](https://gitlab.com/ase/ase/-/blob/master/ase/calculators/orca.py#L46), but at the time the input file is written. The default parameters don't include an `engrad` keyword that turns on the forces calculation. 
    -  Enforce the force calculation by adding an `engrad` keyword. Currently, if it's not supplied by the user, `get_forces()` either fails or runs orca for a second time. 
    - Get default singlet/doublet multiplicity if none is given. 
    - Check convergence. Currently, ase raises no error if "Wavefunction is not converged" is found in the output [here](https://gitlab.com/ase/ase/-/blob/master/ase/io/orca.py#L83).
    - Read geometry optimised structure and optimisation trajectory, if the keyword is supplied by the user (a wfl-only extension)
    - apply a user-supplied post-processing structure (e.g. use an external program to extract charges with natural population analysis)
- Update unit tests. In particular, run the calculation and compare the results to reference values, instead of parsing output from saved orca outputs without running the program itself. 

The documentation example and frequency reading still need updating, but I would like to contribute those in a separate PR, so that the major fixes aren't any more delayed. 

All unit tests pass locally, except for the one marked with an x-fail. 

@bernstei how does this look?


gelzinyte:
No idea why there's a problem with the unit tests' torch set up, locally that bit executes without a problem.

bernstei:
> No idea why there's a problem with the unit tests' torch set up, locally that bit executes without a problem.

It's a pip version issue, which is fixed in #332

bernstei:
@gelzinyte I just merged #330, so maybe merge from main into this branch, so we can confirm there are no bad interactions?

gelzinyte:
> It's a pip version issue, which is fixed in https://github.com/libAtoms/workflow/pull/332 (or, more accurately, should be, but I'm waiting for my suggested fix to be tested)

Ah, missed that. Let's hear from that PR before fixing these tests.

gelzinyte:
> @gelzinyte I just merged https://github.com/libAtoms/workflow/pull/330, so maybe merge from main into this branch, so we can confirm there are no bad interactions?

The local orca tests still pass ok after the merge

bernstei:
@gelzinyte can you merge from `main` again, to see if it fixes the CI torch error and all the tests pass?

PR #334, opened 2024-08-16
Update version

gelzinyte:
Wrap-up version update, once #333  and #330 are merged. 

Some things left to iron out: 

- [ ] Make `pip install wfl==0.2.7` install ASE v3.22.1, not v3.23.0 .
- [ ] Make docs automatically update. With a new PR? With a new tag? With a new version? 
    - [ ] Add option to see docs for different tags?


gelzinyte:
How does the publishing to pip work? I can see that the wheels are built on every push, but does the publishing only happen after a tag? Has to be triggered manually?

bernstei:
If you create a new tag (perhaps has to be of the correct format) it'll push to pypi.  I'm not sure how to retroactively change the requirements for 0.2.7.  Maybe create a branch from 0.2.7, change the requirement in `pyproject.toml` and tag it 0.2.8. Then tag the latest main as 0.3.1?

bernstei:
It's done by https://github.com/libAtoms/workflow/blob/main/.github/workflows/publish_to_pypi.yml, and that has the url with the instructions we followed (and @hheenen did the pypi side).

bernstei:
Looks to me, from that workflow, that any tag triggers a pypi publish.

bernstei:
I'm planning to merge #333 and #336 , assuming they pass the tests, and then we should do this version tag update (at least the v0.3.1 part, so the pypi version is the latest).

bernstei:
@gelzinyte are we ready to increment the internal version number to 0.3.1, then merge it and tag the result as 0.3.1 so pypi gets it?  

We can branch from 0.2.7 (maybe call the branch `0.2.8_final`), do nothing except require a different ASE version and increment the _source code's version_ to 0.2.8, and then git tag that commit as `0.2.8` so pypi gets it.

bernstei:
I merged from main, we'll see if that causes the tests to pass

gelzinyte:
The last thing I wanted to fix was documentation deployment - https://libatoms.github.io/workflow/ hasn't been updated in a while. I'm not sure why `github.event.action == "push"` didn't work. I've copied the deployment action, so docs are also updated on a new tag. I haven't increased the internal version yet, in case you have comments. If not, I think both versions can be increased as you described? Feel free to go ahead or I can do that tomorrow.

gelzinyte:
Ah that shouldn't be there. Removed in the last commit.

bernstei:
OK, tests pass.  I'm going to update the version to 0.3.1, then merge, and tag the resulting `main` version.

PR #340, opened 2024-10-08
Attach neb

jungsdao:
Enable wfl neb.py to have user customized function to be attached to optimizer.
Also allow abort check class to be provided by user.


bernstei:
@jungsdao let me know when it's ready for us to look at.  Could you add a pytest (or modify an existing pytest) to make sure this new capability is used?

Also, a mace descriptors functionality (which I see you added and then removed) would be useful, although perhaps it would be cleaner to do as a separate PR.

PR #351, opened 2025-02-12
fix bug with remote jobs for an op that takes a ConfigSet as its input

### LINKED ISSUE #349: combining couple of operations in one remote submission job
I have a question about wfl package. 
When I want to perform couple of operations for example, minima hopping + optimization or 1st NEB (without climbing image) 2nd NEB (with climbing image), and if I want to do this series of operations by remote submission, is there any way to do it in one submission rather than separate submission?
In current `AutoparaInfo` with `remote_info` takes one remote_info per each operation which makes separate remote submissions is the only way. But more efficient way (especially when one of operation is much cheaper than the other one) would be carrying out all operations in one remote submission.  
I wonder if combining multiple operations into one remote submission is possible. 
I'll be appreciated for any comments on this issue.

bernstei:
Fix _very_ subtle bug in remote jobs with an op that takes a `ConfigSet` as its basic input, rather than an `Atoms` object. 

The subtlety of this bug feels like a bad sign for the overall design, but the tiny fix suggests maybe it's OK

closes #349

### DIFF
--- a/.github/workflows/pytests.yml
+++ b/.github/workflows/pytests.yml
@@ -208,7 +208,12 @@ jobs:
           echo "SLURM apt-get done, doing local configuration"
           host_s=$( hostname -s )
           if [ -f /etc/slurm/slurm.conf ]; then sudo mv /etc/slurm/slurm.conf /etc/slurm/slurm.conf.orig; fi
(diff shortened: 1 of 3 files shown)


bernstei:
Looks like pwscf on these containers is broken, but since the tests pass on my machine, I'm going to merge if that's the only failure, and I'll open a different issue for fixing pw.x.

PR #355, opened 2025-07-10
Allow for more general properties to be saved by the generic calculator

gelzinyte:
In the `main` branch, only the properties defined in `per_atom_properties` or `per_config_properties` are saved by the generic calculator, even if an extra property (e.g., "polarization" for `aims`) is among the calculator's `implemented_properties` and in the `calculator.results` dictionary. The change tries to save all of the properties requested from the calculator and the missing or not implemented properties are caught by `calc.get_property`.

### DIFF
--- a/wfl/utils/save_calc_results.py
+++ b/wfl/utils/save_calc_results.py
@@ -89,7 +89,6 @@ def save_calc_results(atoms, *, prefix, properties):
             config_results['converged'] = atoms.calc.converged
     except AttributeError as exc:
         pass
-        
     if "extra_results" in dir(atoms.calc):
         if prefix is None and (len(atoms.calc.extra_results.get("config", {})) > 0 or
                                        len(atoms.calc.extra_results.get("atoms", {})) > 0):


wfl/utils/save_calc_results.py
```diff
@@ -1,10 +1,11 @@
 import re
+import warnings
 
 from ase.calculators.singlepoint import SinglePointCalculator
 from ase.calculators.calculator import all_properties, PropertyNotImplementedError
 
 per_atom_properties = ['forces', 'stresses', 'charges', 'magmoms', 'energies']
-per_config_properties = ['energy', 'stress', 'dipole', 'magmom', 'free_energy']
```
gelzinyte, line 7: `free_energy` was not read before (no `atoms.get_free energy`). Now it's saved if it's among the properties that were asked to be calculated.

wfl/utils/save_calc_results.py
```diff
@@ -82,6 +83,22 @@ def save_calc_results(atoms, *, prefix, properties):
...
+    # Additional properties asked for in calculation and present in 
+    # .results, but without a specific getter function  
+    remaining_properties = [prop for prop in properties 
+                            if prop not in per_atom_properties 
+                            and prop not in per_config_properties]
+
+    for prop_name in remaining_properties:
+        prop = atoms.calc.get_property(prop_name, allow_calculation=False)
+
+        per_atom = is_per_atom(prop, num_atoms=len(atoms))
```
gelzinyte, line 96: Trying to match the shape and guess whether a property is per-atom is a bit flimsy, but only one of the he `implemented_properties` would have made past `calc.get_property`.

wfl/calculators/aims.py
```diff
@@ -48,8 +48,6 @@ class Aims(WFLFileIOCalculator, ASE_Aims):
     **kwargs: arguments for ase.calculators.aims.Aims
         See https://wiki.fysik.dtu.dk/ase/_modules/ase/calculators/aims.html.
     """
-    implemented_properties = ["energy", "forces", "stress"]
```
gelzinyte, line 51: All of the properties implemented by the parent class are accessible from this calculator, no need to restrict them here.

bernstei: sounds good

wfl/utils/save_calc_results.py
```diff
@@ -113,6 +131,34 @@ def save_calc_results(atoms, *, prefix, properties):
...
+
+    Returns 
+    -------
+    True if shape of the property matches number of atoms 
+    """
+
+    if isinstance(prop, (int, float)):
+        return False
+
+    if num_atoms in [1, 3]:
```
bernstei, line 152: I _really_ don't like this.  This is why ASE has problems with arbitrary properties.

bernstei: I would prefer that, if it's not _too_ cumbersome for us

wfl/utils/save_calc_results.py
```diff
@@ -49,39 +50,26 @@ def save_calc_results(atoms, *, prefix, properties):
...
-    if 'stresses' in properties:
-        atoms_results['stresses'] = atoms.get_stresses()
-    if 'charges' in properties:
-        atoms_results['charges'] = atoms.get_charges()
-    if 'magmoms' in properties:
-        atoms_results['magmoms'] = atoms.get_magnetic_moments()
-    if 'energies' in properties:
-        atoms_results['energies'] = atoms.get_potential_energies()
-
+
```
bernstei, line 72: remove this extra blank line?

wfl/utils/save_calc_results.py
```diff
@@ -99,6 +87,7 @@ def save_calc_results(atoms, *, prefix, properties):
         if "relaxed_positions" in atoms.calc.extra_results:
             atoms.set_positions(atoms.calc.extra_results["relaxed_positions"])
 
+
```
bernstei, line 90: also this blank line

bernstei:
I'm happy with it, although some of this stuff can probably be better handled using https://gitlab.com/ase/ase/-/blob/master/ase/outputs.py?ref_type=heads, as `ase.io.extxyz` does now.  Looks like the names you used, which are presumably from specific calculators you want to use, don't match the standard names in there (e.g. `fermi` vs. `fermi_level`).

gelzinyte:
Let me have a look at outputs.py. And I really am only interested in polarization, for now, so let's skip `fermi` (from vasp) and `charges` (from castep). Do you know why the quantum espresso tests are failing? They run ok for me locally, also when qe is installed with `apt-get`.

bernstei:
There are some issues related to property `free_energy` not being implemented, which could have something to do with your changes.  But many of the QE failures are because the executable gets a SIGABRT, which is weird, but it's not giving any more info as to why.  Let me try to investigate.

bernstei:
I think yes - this is what happens when I run pw.x in the CI

```
Warning: card &CELL ignored
Warning: card / ignored
Warning: card &FCP ignored
Warning: card / ignored
Warning: card &RISM ignored
Warning: card / ignored

     Current dimensions of program PWSCF are:
     Max number of different atomic species (ntypx) = 10
     Max number of k-points (npk) =  40000
     Max angular momentum in pseudopotentials (lmaxx) =  3
*** buffer overflow detected ***: terminated

Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
#0  0x7f10a4423e59 in ???
#1  0x7f10a4422e75 in ???
#2  0x7f10a404532f in ???
#3  0x7f10a409eb2c in pthread_kill
#4  0x7f10a404527d in gsignal
#5  0x7f10a40288fe in abort
#6  0x7f10a40297b5 in ???
#7  0x7f10a4136c18 in __fortify_fail
#8  0x7f10a41365d3 in __chk_fail
#9  0x7f10a4137db4 in __snprintf_chk
#10  0x5557fea0896e in ???
#11  0x5557fe8261d7 in ???
#12  0x5557fe88a007 in ???
#13  0x5557fe70a0c7 in ???
#14  0x5557fe708f22 in ???
#15  0x7f10a402a1c9 in ???
#16  0x7f10a402a28a in __libc_start_main
#17  0x5557fe708f54 in ???
#18  0xffffffffffffffff in ???
Aborted (core dumped)
```

bernstei:
Looks like these failures are probably real side effects of changes you're making.

gelzinyte:
The examples' test was failing because the xtb calculator calculates `free_energy` by default, but it's not among its `implemented_properties`, so `calc.get_property("free_energy")` fails. I have added an extra check for saved properties to avoid this. This should be it, once the tests pass.

PR #359, opened 2025-07-17
switch from calc.get_property to calc.results.get

gelzinyte:
After all, I think this should be `calc.results.get`, not `calc.get_property`. If no `properties` are given, that defaults to all of the standard properties in `calc.results`, so a direct `.get` is more straightforward. That's what happens in `calc.get_property(name, allow_calculation=False)` anyway, but it also checks if the property is in `calc.implemented_properties`. And calculators aren't consistent about that, e.g. `free_energy` is in `calc.results`, but not in `calc.implemented_properties` for xtb. `implemented_properties` don't matter if we use `calc.results.get` instead.

### DIFF
--- a/wfl/utils/save_calc_results.py
+++ b/wfl/utils/save_calc_results.py
@@ -60,20 +60,14 @@ def save_calc_results(atoms, *, prefix, properties):
             if prefix + p in atoms.arrays:
                 del atoms.arrays[prefix + p]
 
-    # Make note of implemented properties, if applicable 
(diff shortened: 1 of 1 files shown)


bernstei:
One possible typo in the error message, otherwise I'm happy with it

gelzinyte:
That's a typo.. Great, thanks for having a look!

bernstei:
I'll merge as soon as the tests pass

PR #365, opened 2026-07-21
optimizer as an argument for optimize

### LINKED ISSUE #364: Specifying which optimizer to be used
Hello, I think it would be nice if user can specify which optimizer to be used for `optimize` function. 
I have some modification which I can try with PR.

jungsdao:
As mentioned in #364, I made small change that can take different optimizer for optimize function because some cases, results are quite sensitive to the optimizer. 
Hope this might be a useful change. 

closes #364


wfl/generate/optimize.py
```diff
@@ -73,6 +73,8 @@ def _run_autopara_wrappable(atoms, calculator, fmax=1.0e-3, smax=None, steps=100
         optimisation logs are not printed unless this is True
     update_config_type: ["append" | "overwrite" | False], default "append"
         whether/how to add at.info['optimize_config_type'] to at.info['config_type']
+    optimizer : ASE optimizer
+        optimizer to use, default LBFGSPrecon
```
bernstei: Can you use the proper ASE function name here?

wfl/generate/optimize.py
```diff
@@ -93,7 +95,7 @@ def _run_autopara_wrappable(atoms, calculator, fmax=1.0e-3, smax=None, steps=100
 
     calculator = construct_calculator_picklesafe(calculator)
 
-    if smax is None:
+    if "precon" in optimizer.__module__.split(".") and smax is None:
```
bernstei: Would it be safer to do `isinstance(optimizer, PreconLBFGS)`?

jungsdao: It seems `isinstance` is not good way to check it. 
```
optimizer = PreconLBFGS
isinstance(optimizer, PreconLBFGS)
## False
``` 
Also there's `PreconFIRE`. I don't have better idea than this but if you have better solution, please let me know I'll adapt it.

wfl/generate/optimize.py
jungsdao: This part has been fixed.

wfl/generate/optimize.py
bernstei: You're right, `isinstance` is wrong. Maybe `optimizer.__name__ in ["PreconLBFGS", "PreconFIRE"]`?  Really it would be better to do this by trying to pass `smax` to the dynamics and falling back if it fails. Let me try to write something like that.

bernstei: Try this diff file 
[opt_smax_diff.txt](https://github.com/user-attachments/files/30244321/opt_smax_diff.txt)

bernstei: Could also write it with nested try/except, but I think that ends up uglier

bernstei: This is the try/catch version 
[opt_smax_diff_trycatch.txt](https://github.com/user-attachments/files/30244476/opt_smax_diff_trycatch.txt)

Neither is well tested

jungsdao: It seems to work for both precon and non-precon optimizers though it's a bit longer. 
There seems to be a test failures which is not related to this PR.

bernstei: Looks like for some reason `gap_fit` doesn't exist in the CI.  I'll merge this PR, and then try to figure out why that's happening.

PR #366, opened 2026-08-14
Updates to support numpy 2

gelzinyte:
Hi!

I'd like to lift the hard `numpy<2` cap, so that Workflow can be used with numpy2-only packages (e.g., [fhi-vibes](https://vibes-developers.gitlab.io/vibes/)). 

Small updates:
- Three calls switched to use numpy-2 syntax
- Changed project dependencies
- Switched the GFN2-xTB backend from the `xtb-python` package (legacy; restricts to numpy<2) to `tblite`

Once this PR is merged, would it also be possible to issue a new PyPi release?

### DIFF
--- a/.github/workflows/pytests.yml
+++ b/.github/workflows/pytests.yml
@@ -19,7 +19,6 @@ jobs:
     env:
       coverage-on-version: "3.10"
       use-mpi: True
-      PIP_CONSTRAINT: pip_constraint.txt
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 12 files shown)


bernstei:
Looks good to me

bernstei:
Looks like tests are failing, maybe on some ASE update changes?  @gelzinyte can you fix those, or should I?

bernstei:
We should also think about ASE compatibility more generally.  For example, `MaxwellBoltzmannDistribution` first switched from `communicator` to `comm`, and is now deprecated.

bernstei:
I agree about ASE, we just need to decide on a policy.  We can basically keep up to date so it definitely works with the latest ASE, and then try to check earlier versions and set the minimum dependency version to something that definitely works.

